PR #8, opened 2020-10-05
COST-546: Get initial upload in place 

abaiken:
## TODO: 
* Add cert verification
* Add tests
## Testing Instructions: 
1. Either run the operator locally or in the cluster from this branch 
2. Deploy the CR with the default token authentication: 
```
make deploy-cr
```
3. Watch the operator logs - reach out to me for a tar.gz (or you can use any tar.gz to upload).
* If you are running locally - create the `/tmp/cost-mgmt-operator-reports/` directory copy the tarfile to that location. 
* If you are running in the cluster you can copy the tarfile to the pod using this command (You will replace the podname with your specific podname): 
```
kubectl cp payload1.tar.gz korekuta-operator-go-controller-manager-5ffd598cbd-678fp:/tmp/cost-mgmt-operator-reports/ -c manager
```
4. Watch the operator logs 
5. Repeat with basic authentication: 
```
make deploy-cr USER=$YOUR_USERNAME PASS=$YOUR_PASSWORD AUTH=basic
```
6. For fun - use an incorrect password to see the error logs for 401 and also check the status of the CR to make sure that the status has all of the correct information. 

Reach out with any questions


go.mod
```diff
@@ -12,6 +12,8 @@ require (
 	k8s.io/api v0.18.2
 	k8s.io/apimachinery v0.18.2
 	k8s.io/client-go v0.18.2
+	k8s.io/component-base v0.18.2
+	k8s.io/klog v1.0.0
```
chambridge, line 16: What requires these dependencies? Not an issue, but couldn't see where they were referenced.

controllers/costmanagement_controller.go
```diff
@@ -262,6 +272,83 @@ func GetAuthSecret(r *CostManagementReconciler, costInput *CostManagementInput,
...
+	}
+	req = req.WithContext(ctx)
+	req.Header.Set("Content-Type", mw.FormDataContentType())
+	if costInput.Authentication == "basic" {
+		log.Info("Uploading using basic authentication!")
+		req.SetBasicAuth(costInput.BasicAuthUser, costInput.BasicAuthPassword)
+	} else {
+		log.Info("Uploading using token authentication")
+		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", costInput.BearerTokenString))
+		req.Header.Set("User-Agent", fmt.Sprintf("cost-mgmt-operator/%s cluster/%s", costInput.OperatorCommit, costInput.ClusterID))
```
chambridge, line 316: Should we just add this no matter what the auth. Can't hurt, right?

api/v1alpha1/costmanagement_types.go
```diff
@@ -159,6 +159,15 @@ type CostManagementStatus struct {
 	// UploadWait is a field of CostManagement to represent the time to wait before sending an upload.
 	UploadWait *int64 `json:"upload_wait,omitempty"`
 
+	// Last Upload Status
+	LastUploadStatus string `json:"last_upload_status,omitempty"`
+
+	// Last Upload Time
+	LastUploadTime string `json:"last_upload_time,omitempty"`
```
chambridge, line 166: Would be nice to additionally capture when the last upload was successful, so you can get an idea for the gap of time and know that you have sent some data.

Dockerfile
```diff
@@ -20,9 +20,20 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
+# FROM gcr.io/distroless/static:nonroot
+FROM alpine:3.7
```
chambridge, line 24: Could try `FROM openshift/origin-release:golang-1.15 as builder` it may have git there. 

https://github.com/openshift/release/blob/65f3976811b3a4fe333339b87a5c8b766c60b775/projects/origin-release/golang-1.15/Dockerfile

chambridge (commented):
Some early feedback

go.mod
abaiken: I couldn't see where either so I ran `go mod tidy` and it removed them. I must have deleted the code that used them!

controllers/costmanagement_controller.go
abaiken: When I add these with basic authentication I get 401 unauthorized. When I removed them with basic auth it went back to 202s.

chambridge: ok. interesting, but I guess it is what it is.

controllers/costmanagement_controller.go
```diff
@@ -262,6 +273,105 @@ func GetAuthSecret(r *CostManagementReconciler, costInput *CostManagementInput,
...
+
+	if resp.StatusCode >= 300 || resp.StatusCode < 200 {
+		body, _ := ioutil.ReadAll(resp.Body)
+		if len(body) > 1024 {
+			body = body[:1024]
+		}
+		log.Info(fmt.Sprintf("gateway server reported unexpected error code: %d (request=%s): %s", resp.StatusCode, requestID, string(body)))
+	}
+
+	if resp.StatusCode == http.StatusAccepted {
```
chambridge, line 361: So recently the ingress service changed its status response code and broke the insights operator. I think it might be "safer" consider >= 200 and < 300 as a success.

abaiken: That's a good call - I will change it!

chambridge (approved):
LGTM

PR #9, opened 2020-10-06
Simplify deployment of dev branches.

chambridge:
* Quick make command update to deploy current branch


abaiken (approved):
This worked for me. I was originally on an older version of git (2.15.1) which did not have the --show-current option but I updated and it worked 👍

PR #10, opened 2020-10-08
Add upload variables for toggling upload/upload cycle 

abaiken:
## Description 
This PR adds the following to the CR spec: 
* UploadCycle - the schedule we are uploading on in minutes - defaults to 360 min
* UploadToggle - a boolean stating whether or not we are uploading - defaults to true

This PR adds the following to the CR status: 
* OperatorCommit - the commit hash of the operator 
* UploadCycle - the schedule we are uploading on in minutes
* UploadToggle - a boolean stating whether or not we are uploading 
* LastUploadStatus - a string with the response code and status
* LastUploadTime - the last attempted upload time
* LastSuccessfulUploadTime - the last successful upload time 

It also does the following: 
* changes to using the ca-bundle.crt for uploading to cloud.redhat.com 
* Adds logic to check the UploadToggle spec field to see if a customer has specified not to upload 
* Adds logic to check and see when the LastSuccessfulUploadTime was, and check whether or not it has been longer/shorter than the specified UploadCycle. 

## Testing:
1. Locally: 
* Deploy the CR to the cluster 
* create the /tmp/cost-mgmt-operator-reports/ directory & copy a tarfile to that location.
* you will need to copy the ca-bundle that the operator creates locally - you can reach out to me for the file if you don't feel like deploying the operator to the cluster first to get the bundle. This ca-bundle.crt file needs to exist at `/var/run/configmaps/trusted-ca-bundle/ca-bundle.crt`

2. Cluster: 
* Deploy the operator and the CR to the cluster
* Watch the operator logs - reach out to me for a tar.gz (or you can use any tar.gz to upload).
If you are running in the cluster you can copy the tarfile to the pod using this command (You will replace the podname with your specific podname):
kubectl cp payload1.tar.gz korekuta-operator-go-controller-manager-5ffd598cbd-678fp:/tmp/cost-mgmt-operator-reports/ -c manager
* Watch the operator logs

## Testing Scenarios: 
* Set the upload_toggle in the CR spec to false - make sure that upload doesn't run (you should see a log that the operator is not configured to upload) 
* Set the upload_toggle to true - make sure that upload attempts to run 
* change the upload_cycle to X min and make sure that once X min has passed since the last successful upload, that upload is ran again. 

## Note 
* I will add unit tests after I merge this PR and get the split payload logic in place so that I can add unit tests for the entire issue functionality


api/v1alpha1/costmanagement_types.go
```diff
@@ -36,6 +36,15 @@ const (
 
 	// CertCheck allows certificate validation to occur.
 	CertCheck bool = true
+
+	//UploadOn sets the operator to upload to cloud.redhat.com
+	UploadOn bool = true
+
+	//UploadOff sets the operator to not upload to cloud.redhat.com
+	UploadOff bool = false
```
maskarb, line 44: do we need both of these?

chambridge: These are constants so I think its ok.

api/v1alpha1/costmanagement_types.go
```diff
@@ -68,6 +77,24 @@ type AuthenticationSpec struct {
...
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	UploadWait *int64 `json:"upload_wait,omitempty"`
+
+	// UploadCycle is a field of CostManagement to represent the number of hours between each upload schedule
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	UploadCycle *int64 `json:"upload_cycle,omitempty"`
+
+	// UploadToggle is a field of CostManagement to represent if the operator should upload to cloud.redhat.com.
```
chambridge, line 93: Maybe mention the default is `true`?

api/v1alpha1/costmanagement_types.go
```diff
@@ -68,6 +77,24 @@ type AuthenticationSpec struct {
...
 
+// UploadSpec defines the desired state of Authentication object in the CostManagementSpec
+type UploadSpec struct {
+
+	// UploadWait is a field of CostManagement to represent the time to wait before sending an upload.
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	UploadWait *int64 `json:"upload_wait,omitempty"`
+
+	// UploadCycle is a field of CostManagement to represent the number of hours between each upload schedule
```
chambridge, line 88: Maybe mention the default is 6 hours? Do we want to do this in hour or minutes (default 3600) to allow us to be able to more easily test the upload cycle?

controllers/costmanagement_controller.go
```diff
@@ -132,17 +135,33 @@ func ReflectSpec(r *CostManagementReconciler, cost *costmgmtv1alpha1.CostManagem
...
 		costInput.UploadWait = r.Int63() % 35
 	}
 
+	if !reflect.DeepEqual(cost.Spec.Upload.UploadCycle, cost.Status.Upload.UploadCycle) {
+		cost.Status.Upload.UploadCycle = cost.Spec.Upload.UploadCycle
+	}
+	if cost.Status.Upload.UploadCycle != nil {
+		costInput.UploadCycle = *cost.Status.Upload.UploadCycle
+	} else {
+		costInput.UploadCycle = 6
```
chambridge, line 162: Lets use the constant here:
https://github.com/project-koku/korekuta-operator-go/pull/10/files#diff-9a43762653cce0c8d43de2353831f3ddR36

chambridge (changes requested):
Minor comments.

api/v1alpha1/costmanagement_types.go
maskarb: Ahh ok, I understand now.

chambridge: We could use one and just do `!UploadOn` which is fine, but then let's do the same for the `CertIgnore`. I'm flexible either way. I think its about readability, but I can see also avoiding anyone ever mistakenly setting UploadOff as `true`.

api/v1alpha1/costmanagement_types.go
abaiken: Added.

api/v1alpha1/costmanagement_types.go
abaiken: Added

chambridge (approved):
LGTM

PR #11, opened 2020-10-09
COST-543 and COST-544- Prometheus functionality

maskarb:
This PR adds:
- establishes a connection to thanos-querier
- performs the necessary queries
- builds the report files and saves them in /tmp/cost-mgmt-operator-reports/data

The queries are run once each hour, collecting the metrics for the previous hour.


collector/collector.go
```diff
@@ -0,0 +1,353 @@
...
+		"namespace-labels":             {"namespace", "kube_namespace_labels"},
+		"node-labels":                  {"node", "kube_node_labels"},
+		"persistentvolume-labels":      {"persistentvolume", "kube_persistentvolume_labels"},           // namespace,persistentvolume,pod
+		"persistentvolumeclaim-labels": {"persistentvolumeclaim", "kube_persistentvolumeclaim_labels"}, // namespace,persistentvolumeclaim,pod
+		"pod-labels":                   {"pod", "kube_pod_labels"},
+		// "pod-persistentvolumeclaim-info": {"pod", "kube_pod_spec_volumes_persistentvolumeclaims_info"},  // not used?
+	}
+)
+
+func DoQuery(promconn prom.API) error {
```
maskarb, line 78: This whole function need a refactor.

abaiken (commented):
This is looking really great so far! I have one comment and then I'll sync up with you about the error I'm seeing when running: 
```
E1012 20:19:20.208639       1 runtime.go:78] Observed a panic: &runtime.TypeAssertionError{_interface:(*runtime._type)(0x1550ec0), concrete:(*runtime._type)(nil), asserted:(*runtime._type)(0x150ea40), missingMethod:""} (interface conversion: interface {} is nil, not string)
```

collector/prometheus.go
```diff
@@ -0,0 +1,224 @@
...
+	err := getRuntimeObj(ctx, r, route, objKey, "route")
+	if err != nil {
+		return err
+	}
+
+	if route.Spec.Host == "" {
+		return fmt.Errorf("getPromAddress: no routes found")
+	}
+	cfg.Address = "https://" + route.Spec.Host
+	return nil
```
chambridge, line 140: Is there a reason why we can't use the service directly instead of the route? If we are running internal to the cluster it doesn't seem necessary to have to use an external route. One of the recent Solution Architects actually called this out in our current operator for how we interact with the report operator and said that this breaks any customer that puts an SSO config in front of any external route.

Additionally, for potential work arounds it might make sense to have a SPEC config item to specify the prometheus address (for testing or other issues).

config/rbac/role.yaml
```diff
@@ -64,6 +65,13 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - route.openshift.io
+  resources:
+  - routes
+  verbs:
+  - get
+  - list
```
chambridge, line 74: This is associated with my route question above.

main.go
```diff
@@ -56,7 +57,7 @@ func init() {
 func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
-	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&metricsAddr, "metrics-addr", ":8081", "The address the metric endpoint binds to.")
```
chambridge, line 60: Not sure I understand this change.

Isn't it tied to the deployment? you can't just change the port in the code, right?
https://github.com/project-koku/korekuta-operator-go/blob/0c17cef1ac360921c9c14e9904fb457fdcd5be11/config/default/manager_auth_proxy_patch.yaml#L24

chambridge (commented):
Few comments about accessing Prometheus from within the cluster.

main.go
maskarb: Yea, this was accidental. I had some issue with 8080 and forgot to change it back.

collector/prometheus.go
maskarb: I can certainly rework this one. I was having some trouble with the svc address (I can't remember why). But I just tried again with `https://thanos-querier.openshift-monitoring.svc:9091` in the cluster, and that worked just fine.

I'll work on a different solution for using an external route for local development.

chambridge: We don't have to block on this if you just want to do a follow up PR.

maskarb: Went ahead and took care of the route.
For local dev, we can add an external address to the CR under the `prometheus_config`:

```
spec:
  prometheus_config:
    address: >-
      https://thanos-querier-openshift-monitoring.apps.cluster....
status:
  authentication:
    credentials_found: true
    type: token
  clusterID: 6534a350-4f43-4393-9383-8e37939e9922
  ingress_url: 'https://cloud.redhat.com/api/ingress/v1/upload'
  operator_commit: f27bf0b8fa5f1aed9744b1b8d51e945fe7b63271
  prometheus:
    address: >-
      https://thanos-querier-openshift-monitoring.apps.cluster...
    last_query_start_time: '2020-10-14T22:39:00Z'
    last_query_success_time: '2020-10-14T22:39:00Z'
    prometheus_connected: true
```

kholdaway (approved):
This is awesome!  Definitely not a 👊 🆙 📈  PR.  Solid work.

PR #12, opened 2020-10-14
Extract out the Http communication with cloud.redhat.com and Config object.

chambridge:
* Restructure some code from the controller
* Create a utils package to hold communication to cloud.redhat.com

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,7 @@ COPY main.go main.go
 COPY api/ api/
 COPY controllers/ controllers/
 COPY clusterversion/ clusterversion/
+COPY crhchttp/ crhchttp/
 
 # Build
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 GO111MODULE=on go build -a -o manager main.go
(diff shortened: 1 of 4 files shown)


utils/config.go
```diff
@@ -0,0 +1,43 @@
...
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+package utils
```
maskarb, line 20: I like moving these out of the controller, but it would be good to make this package more specific. I can foresee `utils` becoming a junk drawer of stuff. Since everything here is related to cloud.redhat, maybe something like `package crhchttp`?

chambridge: Sure. I think that is a true statement in terms of potential junk drawer.

chambridge: rebased and package name updated.

abaiken (approved):
Ran this locally and upload/upload logic still works. LGTM!

PR #13, opened 2020-10-19
Pod not on Node error

maskarb:
This PR fixes an error that arises when a Pod is not associated with a Node:
```
2020-10-19T15:08:24.556-0400    INFO    controllers.CostManagement      writing node results to file    {"costmanagement": "writeResults", "filename": "/tmp/cost-mgmt-operator-reports/data/cm-openshift-node-labels-lookback-202010.csv", "data set": "node"}
E1019 15:08:24.568296   71545 runtime.go:78] Observed a panic: &runtime.TypeAssertionError{_interface:(*runtime._type)(0x21c47e0), concrete:(*runtime._type)(nil), asserted:(*runtime._type)(0x21d0ea0), missingMethod:""} (interface conversion: collector.CSVStruct is nil, not *collector.NodeRow)
```

### DIFF
--- a/collector/collector.go
+++ b/collector/collector.go
@@ -155,7 +155,6 @@ func GenerateReports(promconn promv1.API, ts promv1.Range, log logr.Logger) erro
 
 	// yearMonth is used in filenames
 	yearMonth := ts.Start.Format("200601") // this corresponds to YYYYMM format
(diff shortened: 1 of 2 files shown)


abaiken (approved):
LGTM 👍

PR #15, opened 2020-10-19
Utilize the validate_cert spec boolean and add cert verification

abaiken:
Tested using basic auth with no validation for CI and using token auth/validation for PROD.


chambridge (approved):
LGTM

PR #16, opened 2020-10-19
add service certificate

maskarb:
This PR adds a service certificate which enables TLS verification for the prometheus query requests.

This cert only works for the internal route (i.e. `https://thanos-querier.openshift-monitoring.svc:9091`). Otherwise, we will still get the `x509: certificate signed by unknown authority` error when attempting prometheus queries.


abaiken (approved):
LGTM!

PR #17, opened 2020-10-20
COST 395 - Create a source for the cluster via the operator.

chambridge:
* Enable mechanism to simply check for the source periodically
* Enable flow to create the source
* Provide status information regarding the state of the source

Examples:

If you want to check if your source with a given name has been registered:
```
spec:
  source:
    name: ocp-dev-cluster
```
Resulting status if the source is not already registered.
```
status:
  api_url: 'https://cloud.redhat.com'
  authentication:
    credentials_found: true
    type: token
  clusterID: 83217c64-9ddf-45b8-be96-35c662cbbfb9
  operator_commit: dd229ede1cc1890b126a12acbf5ba16fd4bc0406
  source:
    error: >-
      No OpenShift source registered with name ocp-dev-cluster and Cluster ID
      83217c64-9ddf-45b8-be96-35c662cbbfb9.
    last_check_time: '2020-10-20T18:31:45Z'
    name: ocp-dev-cluster2
    source_defined: false
    sources_path: /api/sources/v1.0/
```

If you want to create a source with a given name:
```
spec:
  source:
    name: ocp-dev-cluster
    create_source: true
```
Resulting status if the source is not already registered.
```
status:
  api_url: 'https://cloud.redhat.com'
  authentication:
    credentials_found: true
    type: token
  clusterID: 83217c64-9ddf-45b8-be96-35c662cbbfb9
  operator_commit: dd229ede1cc1890b126a12acbf5ba16fd4bc0406
  source:
    last_check_time: '2020-10-20T17:21:15Z'
    name: ocp-dev-cluster
    source_defined: true
    sources_path: /api/sources/v1.0/
```
image

If you try and create a source for a cluster that is already registered with a different name:
```
spec:
  source:
    name: ocp-dev-cluster3
    create_source: true
```
Resulting status if the source is already registered.
```
status:
  api_url: 'https://cloud.redhat.com'
  authentication:
    credentials_found: true
    type: token
  clusterID: 83217c64-9ddf-45b8-be96-35c662cbbfb9
  operator_commit: dd229ede1cc1890b126a12acbf5ba16fd4bc0406
  source:
    error: >-
      OpenShift source with Cluster ID 83217c64-9ddf-45b8-be96-35c662cbbfb9 is
      registered with different name. The cluster may already be registered with
      a different name.
    last_check_time: '2020-10-20T18:36:42Z'
    name: ocp-dev-cluster3
    source_defined: false
    sources_path: /api/sources/v1.0/
```


chambridge:
rebased on latest master.

crhchttp/http_cloud_dot_redhat.go
```diff
@@ -113,21 +116,49 @@ func getClient(logger logr.Logger, validateCert bool) http.Client {
...
+		}
+		log.Info(fmt.Sprintf("Error Response Body: %s", string(body)))
+		return nil, fmt.Errorf(string(body))
+	}
+
+	if resp.StatusCode >= 200 && resp.StatusCode < 300 {
+		log.Info(fmt.Sprintf("Successfully request x-rh-insights-request-id=%s", requestID))
+		bodyBytes, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			log.Error(err, "The following error occurred")
```
maskarb, line 144:
```suggestion
			log.Error(err, "Error occurred reading the response body.")
```

sources/handler.go
```diff
@@ -0,0 +1,487 @@
...
+	lastCheckStr := lastCheck.UTC().Format("2006-01-02 15:04:05")
+	log.Info("The last source check took place at " + lastCheckStr)
+	// transforming the string into a time.Time object
+	checkTime, err := time.Parse("2006-01-02 15:04:05", lastCheckStr)
+	if err != nil {
+		return true
+	}
+	duration := time.Since(checkTime)
+	if int64(duration.Minutes()) >= cycle {
+		log.Info("Checking source oncloud.redhat.com!")
```
maskarb, line 136:
```suggestion
		log.Info("Checking source on cloud.redhat.com!")
```

maskarb (changes requested):
I think we need a small change around how we set `costConfig.LastSourceCheckTime`, but otherwise things look good.

controllers/costmanagement_controller.go
```diff
@@ -147,13 +150,28 @@ func ReflectSpec(r *CostManagementReconciler, cost *costmgmtv1alpha1.CostManagem
...
+		sourceCycleChange = true
+	}
+	if cost.Status.Source.CheckCycle != nil {
+		costConfig.SourceCheckCycle = *cost.Status.Source.CheckCycle
+	} else {
+		costConfig.SourceCheckCycle = costmgmtv1alpha1.DefaultSourceCheckCycle
+	}
+	if sourceNameChanged || sourceCycleChange {
+		costConfig.LastSourceCheckTime = cost.Status.Source.LastSourceCheckTime
+	}
```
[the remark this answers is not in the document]
chambridge: yeah the intention was the flip of this 🤦  if these items were false. My screw up with the last minute change to better handle CR updates. Will fix this.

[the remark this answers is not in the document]
chambridge: updated.

sources/handler.go
```diff
@@ -0,0 +1,487 @@
...
+}
+
+// ApplicationTypeResponse A data structure for the paginated application type response
+type ApplicationTypeResponse struct {
+	Meta GenericMeta
+	Data []ApplicationTypeDataItem
+}
+
+// CheckSource Determine if the source should be checked for existence or creation
+func CheckSource(logger logr.Logger, costConfig *crhchttp.CostManagementConfig) bool {
```
abaiken, line 109: Since this function/check is so similar to the check cycle function (https://github.com/project-koku/korekuta-operator-go/blob/master/controllers/costmanagement_controller.go#L287) is there anyway that we could create one function to use for both situations? I realize that this one is more nuanced in that it checks for the source name but I think we could possibly work around that.

crhchttp/http_cloud_dot_redhat.go
```diff
@@ -113,21 +116,49 @@ func getClient(logger logr.Logger, validateCert bool) http.Client {
...
+		}
+		return bodyBytes, nil
+	}
+	return nil, fmt.Errorf("Unexpected Response")
+}
+
 // Upload Send data to cloud.redhat.com
-func Upload(logger logr.Logger, costConfig *CostManagementConfig, method string, path string, body *bytes.Buffer, mw *multipart.Writer) (string, metav1.Time, error) {
+func Upload(logger logr.Logger, costConfig *CostManagementConfig, method string, uri string, body *bytes.Buffer, mw *multipart.Writer) (string, metav1.Time, error) {
 	log := logger.WithValues("costmanagement", "Upload")
```
abaiken, line 154: Not a blocker and if we wanted to do this it could be in a separate PR - but it might be good to think about the logging in the upload functions that we are using for both ingress and creating sources. For example: 
```
2020-10-20T15:36:31.820-0400    INFO    controllers.CostManagement      OpenShift source type is 1      {"costmanagement": "SourceGetOrCreate"}
2020-10-20T15:36:31.820-0400    INFO    controllers.CostManagement      Configured to not upload using the certificate! {"costmanagement": "GetClient"}
2020-10-20T15:36:31.820-0400    INFO    controllers.CostManagement      Uploading using token authentication    {"costmanagement": "SetupRequest"}
```
The above could be changed to not specifically say upload/Uploading since that could get a little muddled when its being used for creating sources.

abaiken (commented):
Looking great! Non-blocking comments for thought.

crhchttp/http_cloud_dot_redhat.go
chambridge: Pushed a clean up for the logging in these common functions.

controllers/costmanagement_controller.go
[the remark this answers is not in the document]
maskarb: Things are looking better, but we need another check somewhere. I tried creating a source, but I failed authentication. The source was not created, but because of the above change, the time _was_ set, and `CheckSource` return `false`. Maybe we need additional logic in `CheckSource` that looks at spec `create_source` and status `source_defined` and returns `true` if `create=true` and `defined=false`?

sources/handler.go
chambridge: done.

controllers/costmanagement_controller.go
```diff
@@ -284,28 +302,28 @@ func GetAuthSecret(r *CostManagementReconciler, costConfig *crhchttp.CostManagem
...
-	log := r.Log.WithValues("costmanagement", "checkCycle")
-	if !lastSuccess.IsZero() {
+func checkCycle(logger logr.Logger, cycle int64, lastExecution metav1.Time, action string) bool {
+	log := logger.WithValues("costmanagement", "checkCycle")
+	if !lastExecution.IsZero() {
 		// transforming the metav1.Time object into a string
-		lastSuccess := lastSuccess.UTC().Format("2006-01-02 15:04:05")
-		log.Info("The last successful upload took place at " + lastSuccess)
+		lastExecution := lastExecution.UTC().Format("2006-01-02 15:04:05")
+		log.Info("The last successful upload took place at " + lastExecution)
```
maskarb, line 310:
```suggestion
		log.Info(fmt.Sprintf("The last successful %s took place at %s.", action, lastExecution))
```

maskarb (commented):
One other log change

controllers/costmanagement_controller.go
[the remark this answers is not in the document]
chambridge: What happens if you delete the `source` section from the spec save it then add it back. That should force the name to go from a value, to nil, to a value and trigger the source check.

If you get a 401 or 403 I don't think we would want it to keep trying just because `create=true".  Maybe that is not a great experience 🤷 ?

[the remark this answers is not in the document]
maskarb: hmm, that's a good point. I'll give that a try to see what happens.

And yes, thinking about it now, if we're hitting auth errors, it would not be good to watch this loop over and over trying to create something that will never be successful without intervention.

controllers/costmanagement_controller.go
```diff
@@ -147,13 +150,28 @@ func ReflectSpec(r *CostManagementReconciler, cost *costmgmtv1alpha1.CostManagem
...
+	if !reflect.DeepEqual(cost.Spec.Source.CheckCycle, cost.Status.Source.CheckCycle) {
+		cost.Status.Source.CheckCycle = cost.Spec.Source.CheckCycle
+		sourceCycleChange = true
+	}
+	if cost.Status.Source.CheckCycle != nil {
+		costConfig.SourceCheckCycle = *cost.Status.Source.CheckCycle
+	} else {
+		costConfig.SourceCheckCycle = costmgmtv1alpha1.DefaultSourceCheckCycle
+	}
+	if !sourceNameChanged || !sourceCycleChange {
```
maskarb, line 170:
```suggestion
	if !sourceNameChanged && !sourceCycleChange {
```
This needs to be an `and`, not an `or`.

maskarb (commented):
Ok, last change, and it'll be good to go.

maskarb (approved):
Looks good to me!

PR #18, opened 2020-10-21
add local dev doc

maskarb:


README.md
```diff
@@ -14,6 +14,8 @@ To submit an issue please visit https://issues.redhat.com/projects/COST/
 
 ## Development
 
+More specific local development can be found [here](docs/local-development.md).
```
abaiken, line 17:
```suggestion
More specific documentation for local development can be found [here](docs/local-development.md).
```

docs/local-development.md
```diff
@@ -0,0 +1,73 @@
...
+## Pre-reqs
+
+* Access to an Openshift cluster
+* A clone of [korekuta-go-operator](https://github.com/project-koku/korekuta-operator-go)
+* [Go 1.13 or greater](https://golang.org/doc/install)
+* [Openshift-CLI](https://docs.openshift.com/container-platform/4.5/cli_reference/openshift_cli/getting-started-cli.html) (preferably a version that matches your Openshift cluster version)
+* [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
+* [kustomize](https://kubernetes-sigs.github.io/kustomize/installation/) (before installing this separately, check that it was not already installed along with kubectl)
+* [Docker Desktop](https://www.docker.com/products/docker-desktop)
+* quay.io account
```
abaiken, line 12: could add a link to quay.io here since you did for the other things

docs/local-development.md
```diff
@@ -0,0 +1,74 @@
+# Local Development
+
+## Pre-reqs
+
+* Access to an Openshift cluster
```
abaiken, line 5: should we say 4.3 or greater here?

docs/local-development.md
```diff
@@ -0,0 +1,74 @@
...
+    The operator is running but is not doing any work. We need to create a CR.
+
+5. Deploy a CR. For now, use basic authentication. The following creates the appropriate spec. `username` and `password` correspond to the username (not email address) and password for the account you want to use at cloud.redhat.com:
+
+    ```
+    $ make deploy-cr AUTH=basic USER=<username> PASS=<password>
+    ```
+    This command uses the CR defined in `config/samples/cost-mgmt_v1alpha1_costmanagement.yaml`, adds the authentication spec, and creates a CR in `testing/cost-mgmt_v1alpha1_costmanagement.yaml`. The command then deploys this CR to the cluster.
+
+    After this CR has been created in the cluster, reconciliation will begin. It will error because the URL for prometheus needs to be set in the CR.
```
abaiken, line 55: It would be nice if we could grab the prometheus URL out of the cluster and add it to the CR in this make command so that we didn't error instead of documenting that it errors.

docs/local-development.md
```diff
@@ -0,0 +1,74 @@
...
+    make run ENABLE_WEBHOOKS=false
+    ```
+
+    At this point, you will see the operator spin up in your terminal. After a few seconds, you should see something similar to the following output:
+    ```
+    2020-10-21T09:31:37.195-0400    INFO    controller-runtime.controller   Starting workers        {"controller": "costmanagement", "worker count": 1}
+    ```
+    The operator is running but is not doing any work. We need to create a CR.
+
+5. Deploy a CR. For now, use basic authentication. The following creates the appropriate spec. `username` and `password` correspond to the username (not email address) and password for the account you want to use at cloud.redhat.com:
```
abaiken, line 48: We should document that if you just run `make deploy-cr` it will use token authentication.

docs/local-development.md
```diff
@@ -0,0 +1,74 @@
...
+
+    ```
+    $ make deploy-cr AUTH=basic USER=<username> PASS=<password>
+    ```
+    This command uses the CR defined in `config/samples/cost-mgmt_v1alpha1_costmanagement.yaml`, adds the authentication spec, and creates a CR in `testing/cost-mgmt_v1alpha1_costmanagement.yaml`. The command then deploys this CR to the cluster.
+
+    After this CR has been created in the cluster, reconciliation will begin. It will error because the URL for prometheus needs to be set in the CR.
+
+6. In the cluster dashboard, under `Networks`, select `Routes`. Change the project to `openshift-monitoring` and copy the route for `thanos-querier`. In `testing/cost-mgmt_v1alpha1_costmanagement.yaml`, add the route and turn off TLS verification:
```
abaiken, line 57: I feel like we could have an oc command get this for us and then add it to the CR in our makefile instead of documenting this. It seems redundant to have developers deploy a CR that will not work for them and then go back to edit and re-deploy it.

docs/local-development.md
```diff
@@ -0,0 +1,74 @@
...
+            service_address: https://thanos-querier-openshift-monitoring.apps.cluster-....
+            skip_tls_verification: true
+    ```
+    Redeploy the CR:
+    ```
+    $ testing/cost-mgmt_v1alpha1_costmanagement.yaml
+    ```
+    Once the CR is updated in the cluster, reconciliation will happen automatically. Now prometheus should be reachable.
+
+7. To continue development, make code changes. To apply those changes, stop the operator, and redeploy it. If changes are made to the api, the CRD needs to be re-registered, and the operator re-deployed.
```
abaiken, line 74: After this, you will still get errors if you do not have a copy of the service cert locally I think.

abaiken (changes requested):
A few minor comments

docs/local-development.md
maskarb: Only when TLS verification is not disabled.

Makefile
```diff
@@ -73,11 +75,20 @@ setup-auth:
...
 
+add-prom-route:
+	@echo 'spec:' >> testing/cost-mgmt_v1alpha1_costmanagement.yaml
+	@echo '  prometheus_config:' >> testing/cost-mgmt_v1alpha1_costmanagement.yaml
+	@echo '    service_address: $(EXTERNAL_PROM_ROUTE)'  >> testing/cost-mgmt_v1alpha1_costmanagement.yaml
+	@echo '    skip_tls_verification: true' >> testing/cost-mgmt_v1alpha1_costmanagement.yaml
+
 deploy-cr:
 	@cp config/samples/cost-mgmt_v1alpha1_costmanagement.yaml testing/cost-mgmt_v1alpha1_costmanagement.yaml
+	$(MAKE) add-prom-route
```
abaiken, line 86: can we add a variable here for whether or not we should add the external route to the spec? Kind of the same way that we check for auth below. I use the `deploy-cr` command whether I'm running as a pod inside the cluster or on my local machine.

docs/local-development.md
abaiken: True - but isn't `SkipTLSVerification` set to false by default? If so, unless a developer manually edits the CR they will be met with the file not found errors.

docs/cr-example.md
```diff
@@ -0,0 +1,28 @@
...
+    upload_toggle: bool # default=true, turn upload on or off -> true means upload, false means do not upload
+  prometheus_config:
+    service_address: string # default=https://thanos-querier.openshift-monitoring.svc:9091, route to thanos-querier
+    skip_tls_verification: bool # default=false, do TLS verification for prometheus queries
+  source:
+    sources_path: string # default=/api/sources/v1.0/, path to sources API
+    name: string # name of source in cloud.redhat.com
+    create_source: bool # default=false, create the source or not
+    check_cycle: int # default=1440, time in minutes to wait between source checks.
+```
```
abaiken, line 28: can we add: 
```
packaging: 
    max_size: int # default=100, min=1, max=100, max number of megabytes for report files. 
```

abaiken (commented):
Looks great!

abaiken (approved):

PR #19, opened 2020-10-21
Report fixes

maskarb:
- fix volume queries. Using `max` instead of `sum` aligns the volume request/capacity metrics with the old operator.
- PV name, labels, storageclassname now come from one query. This covers a case where a PVC/PV exists, but is not associated with a Pod.


abaiken (approved):
LGTM

PR #21, opened 2020-10-26
add file directory to spec

maskarb:
This PR adds a file directory to the spec. This allows us to define a directory for local dev.

### DIFF
--- a/controllers/costmanagement_controller.go
+++ b/controllers/costmanagement_controller.go
@@ -60,8 +60,6 @@ var (
 	authSecretPasswordKey    = "password"
 
 	queryDataDir = "data/"
-	uploadDir    = "upload/"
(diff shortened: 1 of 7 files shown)


controllers/costmanagement_controller.go
```diff
@@ -454,47 +457,43 @@ func (r *CostManagementReconciler) Reconcile(req ctrl.Request) (ctrl.Result, err
 		}
 	}
 
-	upload := checkCycle(r.Log, costConfig.UploadCycle, costConfig.LastSuccessfulUploadTime, "upload")
+	log.Info("Getting Directory Configuration.")
```
abaiken, line 460:
```suggestion
	log.Info("Getting directory configuration.")
```

PR #26, opened 2020-11-02
COST-543 prometheus unittests

maskarb:


collector/prometheus_test.go
```diff
@@ -0,0 +1,625 @@
...
+				t.Errorf("%s got unexpected error: %v", tt.name, got)
+			}
+			if got == nil && tt.err != nil {
+				t.Errorf("%s expected error, got %v", tt.name, got)
+			}
+		})
+	}
+}
+
+var _ = Describe("Collector Tests", func() {
```
abaiken, line 526: this is cool! Still wrapping my head around this haha!

collector/report_test.go
```diff
@@ -0,0 +1,359 @@
...
+	}
+	if err := os.Chmod(tempDir, mode); err != nil {
+		t.Fatalf("Failed to change permissions of temp file.")
+	}
+	return tempDir
+}
+
+func TestWriteReport(t *testing.T) {
+	tempDir := getTempDir(t, os.ModePerm, ".", "test_dir")
+	// tempFile := getTempFile(t, 0777, tempDir)
```
abaiken, line 79: can remove this

collector/report_test.go
```diff
@@ -0,0 +1,359 @@
...
+	tempBadFile := getTempFile(t, 0777, tempDir)
+	tempBadFile.Close()
+	// defer tempFile.Close()
```
abaiken, line 82: can remove this comment too

collector/report_test.go
```diff
@@ -0,0 +1,359 @@
...
+			name: "failed to write to file",
+			report: &report{
+				file: &fakeFile{
+					file:      getTempFile(t, 0777, tempDir),
+					created:   false,
+					createErr: nil,
+				},
+				data: &fakeData{writeErr: errTest},
+			},
+			want: errTest,
```
abaiken, line 136: same comment about making this a bool instead of an error type

collector/report_test.go
```diff
@@ -0,0 +1,359 @@
...
+		},
+		{
+			name: "existing file with no permissions",
+			report: &file{
+				name: tempFileNoPerm.Name(),
+				path: tempDir,
+			},
+			wantedFile:    "",
+			wantedCreated: true,
+			err:           errTest,
```
abaiken, line 216: same comment about err type vs bool.

abaiken (commented):
Looking really good! Here's some minor comments. We should also update the docs to say that kubebuilder is required.

.gitignore
```diff
@@ -34,6 +34,7 @@ commit
 
 # Ignore the testing directory for developing
 testing/*
+collector/test_files/test_reports/*
```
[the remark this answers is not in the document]
maskarb: Couple things: 1. the test expects this folder to exist; fewer things to worry about going wrong for the test if the directory already exists. 2. Everything in this folder is supposed to be cleaned-up, but test failures can prevent cleanup from completing.

collector/prometheus_test.go
```diff
@@ -0,0 +1,625 @@
...
+	}
+	for _, tt := range performMatrixQueryTests {
+		t.Run(tt.name, func(t *testing.T) {
+			col.PromConn = mockPrometheusConnection{
+				singleResult: tt.queryResult,
+				t:            t,
+			}
+			got := mappedResults{}
+			err := col.getQueryResults(&querys{query{QueryString: "fake-query"}}, &got)
+			if err != nil && tt.err == nil {
```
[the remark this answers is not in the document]
maskarb: removed

collector/prometheus_test.go
```diff
@@ -0,0 +1,625 @@
...
+		},
+		{
+			name:        "return incorrect type (model.Vector)",
+			queryResult: &mockPromResult{value: &model.Vector{}},
+			err:         errTest,
+		},
+		{
+			name:        "return incorrect type (model.String)",
+			queryResult: &mockPromResult{value: &model.String{}},
+			err:         errTest,
```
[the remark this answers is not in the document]
maskarb: To me, it makes more sense for the check to read as: "If the expected error is not nil and the gotten error is nil -> test failure" and "If the expected error is nil and the gotten error is not nil -> test failure". It's not a direct comparison of the error strings, but this comparison makes more sense to me than to switch to using a bool.

collector/prometheus_test.go
```diff
@@ -0,0 +1,625 @@
...
+				t.Errorf("%s got unexpected error: %v", tt.name, err)
+			}
+			if tt.err != nil && err == nil {
+				t.Errorf("%s got `%v` error, wanted error", tt.name, err)
+			}
+		})
+	}
+}
+
+func TestGetQueryResults(t *testing.T) {
```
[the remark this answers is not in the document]
maskarb: Because this test has such a big setup, I've kept this one on its own. I left the table setup in case we ever want to add additional test cases.

collector/prometheus_test.go
```diff
@@ -0,0 +1,625 @@
...
+				t.Errorf("%s got: nil error, want: error", tt.name)
+			}
+			if !reflect.DeepEqual(got, tt.wantedResult) {
+				t.Errorf("getQueryResults got:\n\t%s\n  want:\n\t%s", got, tt.wantedResult)
+			}
+		})
+	}
+}
+
+func TestGetQueryResultsError(t *testing.T) {
```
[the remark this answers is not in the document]
maskarb: I consolidated the `TestGetQueryResultsWrongType` into this test.

collector/report_test.go
maskarb: removed

collector/report_test.go
maskarb: removed

abaiken (approved):
👍  LGTM

PR #33, opened 2020-11-19
COST-395 Sources unittests

maskarb:
title says it all!

### DIFF
--- a/sources/handler_test.go
+++ b/sources/handler_test.go
@@ -32,23 +32,42 @@ import (
 )
 
 var (
-	cost       = &crhchttp.CostManagementConfig{Log: testLogger}
+	cost = &crhchttp.CostManagementConfig{
+		APIURL:         "https://ci.cloud.redhat.com",
+		SourcesAPIPath: "/api/sources/v1.0/",
+		Log:            testLogger,
+	}
 	errSources = errors.New("test error")
 	testLogger = logr.NullLogger{}
 )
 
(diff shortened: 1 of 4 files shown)


sources/handler_test.go
```diff
@@ -19,17 +19,1103 @@ along with this program.  If not, see <https://www.gnu.org/licenses/>.
...
+			got, _, err := SourceGetOrCreate(&c, &tt.clts)
+			if tt.expectedErr != nil && err == nil {
+				t.Errorf("%s expected error, got: %v", tt.name, err)
+			}
+			if tt.expectedErr == nil && err != nil {
+				t.Errorf("%s got unexpected error: %v", tt.name, err)
+			}
+			if tt.want != got {
+				t.Errorf("%s got %v want %v", tt.name, got, tt.want)
+			}
```
abaiken, line 1118:
```suggestion
			switch {
			case tt.expectedErr != nil && err == nil:
				t.Errorf("%s expected error, got: %v", tt.name, err)
			case tt.expectedErr == nil && err != nil:
				t.Errorf("%s got unexpected error: %v", tt.name, err)
			case tt.want != got:
				t.Errorf("%s got %v want %v", tt.name, got, tt.want)
			}
```

abaiken (approved):
Looks good! You could optionally clean up the test validation if statements by using switch cases like in the example below but not a blocker.

sources/handler_test.go
maskarb: Discussed offline. Decided to keep as-is.

PR #34, opened 2020-11-24
Add unit tests to dirconfig

boaz0:


dirconfig/dirconfig_test.go
```diff
@@ -0,0 +1,262 @@
...
+		if tc.expectedError != nil || err != nil {
+			if (tc.expectedError.Error() != err.Error()) || (tc.expectedError != nil && err == nil) || (tc.expectedError == nil && err != nil) {
+				t.Errorf("Expected to return error: %v but got %v", tc.expectedError, err)
+			}
+		}
+	}
+}
+
+func TestDirExists(t *testing.T) {
+	dirPath := "/tmp/blalal23392910302919304029191001111fake-non-exist-void-234321123455fd344fdhshjka"
```
maskarb, line 136: Let's make this string random every time the test runs. You can use this helper to create random strings:
[https://github.com/project-koku/korekuta-operator-go/blob/master/collector/suite_test.go#L155](https://github.com/project-koku/korekuta-operator-go/blob/master/collector/suite_test.go#L155)

maskarb (changes requested):
Overall this looks pretty good. Just one change, and then I think we will be good to go: The final return statement in `getOrCreatePath` is not covered. We should have a test for that return statement.

One other comment:

boaz0:
ping @maskarb - can you review this again?

testutils/random_string.go
```diff
@@ -0,0 +1,21 @@
+package testutils
```
maskarb, line 1: This looks good to me. Could you also move the `testlogr.go` into this package and update the references throughout the project? It would be good to minimize the number of test related packages we create.

maskarb (commented):
One more thing, and then this'll be good to go.

testutils/random_string.go
boaz0: Sure!

maskarb (dismissed):
Lgtm! Thanks for taking care of these tests!

maskarb (approved):

maskarb (approved):

PR #35, opened 2020-12-01
check if prometheus spec updated

maskarb:
Testing:
1. Apply a CR with an invalid prometheus service address.
2. Watch reconciler fail to connect to prometheus.
3. Correct the service address and re-apply the CR.
4. Watch connection to prometheus succeed.


abaiken (approved):
LGTM 👍

PR #36, opened 2020-12-02
Refactor controller

abaiken:
This PR does the following: 

- adds CR defaults for the fields that make sense 
- removes the nil check logic since we expect the CR to have the default values 
- removes the costConfig and creates an authConfig and sSpec to simplify/split out default logic 
- breaks the reconciler into logical functions 
- updates the controller tests to work with the above changes 
- injects git commit as an env var rather than file 
- adds logging for the requests headers/ cleans up credential logging so we are not exposing sensitive info


maskarb (approved):
lgtm!

PR #37, opened 2020-12-03
COST-547 add pvc config

maskarb:


abaiken (approved):
This worked. To test, I deployed to the cluster and set my CR to upload to CI ( so that it would fail) - the operator created a tar file called `cost-mgmt-3621ad49-1f4f-481b-a1ef-334cf137e9c5.tar.gz`, I deleted the pod, and watched the new pod attempt to upload the same file, proving that the file was persisted.

PR #38, opened 2020-12-03
COST-667 cost-mgmt -> koku-metrics

maskarb:
This PR changes the name of CostManagement to KokuMetricsConfig to prevent name collisions when installed next to the old operator.


config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
```
abaiken, line 3: I don't think these changes should be committed (this is the file that is supposed to be git ignored but acts weird)

config/crd/bases/koku-metrics.openshift.io_kokumetrics.yaml
```diff
@@ -6,20 +6,20 @@ metadata:
...
+  name: kokumetrics.koku-metrics.openshift.io
 spec:
-  group: cost-mgmt.openshift.io
+  group: koku-metrics.openshift.io
   names:
-    kind: CostManagement
-    listKind: CostManagementList
-    plural: costmanagements
-    singular: costmanagement
+    kind: KokuMetrics
```
abaiken, line 13: I don't think the kind should be plural because then the CRD shows up as a `KokuMetrics` I think we should name it `KokuMetricConfig` or something similar.

config/manager/kustomization.yaml
maskarb: I added a snippet to the readme to prevent tracking this file locally.

config/crd/bases/koku-metrics.openshift.io_kokumetrics.yaml
maskarb: Changed it to `KokuMetricsConfig`

crhchttp/http_cloud_dot_redhat.go
```diff
@@ -105,7 +104,7 @@ func SetupRequest(authConfig *AuthConfig, contentType, method, uri string, body
 	default:
 		log.Info("Request using token authentication")
 		req.Header.Set("Authorization", fmt.Sprintf("Bearer %s", authConfig.BearerTokenString))
-		req.Header.Set("User-Agent", fmt.Sprintf("cost-mgmt-operator/%s cluster/%s", authConfig.OperatorCommit, authConfig.ClusterID))
+		req.Header.Set("User-Agent", fmt.Sprintf("koku-mertics-operator/%s cluster/%s", authConfig.OperatorCommit, authConfig.ClusterID))
```
abaiken, line 107: This cannot be changed unless we submit a PR to the uhc-auth-proxy repo to change our prefix name https://github.com/RedHatInsights/uhc-auth-proxy/pull/5

collector/prometheus.go
```diff
@@ -38,16 +38,16 @@ import (
...
 
 var (
-	promSpec *costmgmtv1alpha1.PrometheusSpec
+	promSpec *kokumetricscfgv1alpha1.PrometheusSpec
 
-	costMgmtNamespace  = "openshift-cost"
-	secretKey          = "token"
-	serviceAccountName = "default"
-	tokenRegex         = "default-token-*"
+	kokuMetricsCfgNamespace = "openshift-cost"
```
abaiken, line 47: My only other comment would be we should discuss if we want to change the namespace for our community operator. I'm not sure if it matters but regardless we can do it in a different PR if we decide to!

crhchttp/http_cloud_dot_redhat.go
maskarb: reverted back to `cost-mgmt-operator`

abaiken (approved):
LGTM

collector/prometheus.go
maskarb: converted namespace to `koku-metrics-operator`

PR #39, opened 2020-12-07
Fix dev make commands and add controller tests 

abaiken:
This PR does the following: 
* fixes the broken makefile commands and makes the deploy commands easier 
* fixes a bug found during unit testing with the upload_wait in the controller 
* adds tests to bump controller coverage 
* updates the readme with the new makefile commands


maskarb (approved):
LGTM!

PR #43, opened 2020-12-09
Change validate_cert default to true

abaiken:
This PR : 
* changes the default validate_cert to true 
* fixes the make command for deploying the local CR to make it false


maskarb (approved):
LGTM

PR #46, opened 2020-12-09
drop configmap in favor of pulling certs from secrets

maskarb:
This PR drops the dependency of a ConfigMap for certificates. The operator already creates several Secrets which contain the necessary certificates. This PR leverages those certificates for connecting to Prometheus. Connecting to cloud.redhat.com relies on certificates that come with the distroless image.

A lot of this code can probably be cleaned up in a future PR. Currently there is repetitive code in the controller and collector/prometheus that can probably be consolidated into a separate package. For now though, this PR drops a dependency that was preventing us from building a release.


abaiken (approved):
LGTM!

PR #47, opened 2020-12-10
Add cloud-dot-redhat tests

boaz0:


boaz0:
What? How does it decrease code coverage :scream:

PR #48, opened 2020-12-10
COST-668 - create bundle

maskarb:
- creates the v0.9.0 bundle.
- makes changes to the Makefile so that bundle creation is completely automated. The only thing that needs to be changed in the future is the version number in the Makefile
- Added some comments on the CR fields that should be used for dev only


config/samples/koku-metrics-cfg_v1alpha1_kokumetricsconfig.yaml
```diff
@@ -3,8 +3,15 @@ kind: KokuMetricsConfig
 metadata:
   name: kokumetricscfg-sample
 spec:
-  authentication: {}
-  packaging: {}
+  authentication:
```
abaiken, line 6: changing this sample will require us to update our make commands for local development.

koku-metrics-operator/0.9.0/manifests/koku-metrics-operator.clusterserviceversion.yaml
```diff
@@ -0,0 +1,262 @@
...
+  namespace: koku-metrics-operator
+spec:
+  apiservicedefinitions: {}
+  customresourcedefinitions:
+    owned:
+    - description: KokuMetricsConfig is the Schema for the kokumetricsconfig API
+      kind: KokuMetricsConfig
+      name: kokumetricsconfigs.koku-metrics-cfg.openshift.io
+      version: v1alpha1
+  description: INSERT OPERATOR DESCRIPTION
```
abaiken, line 54: We should sed replace this with the csv-description docs

maskarb: Instead of sed replace, I created a python script to make all the necessary file changes.

config/samples/koku-metrics-cfg_v1alpha1_kokumetricsconfig.yaml
maskarb: Moved a CR template into the testing file. Now we can make changes to this sample without breaking the Makefile.

Makefile
```diff
@@ -246,10 +242,16 @@ KUSTOMIZE=$(shell which kustomize)
 endif
 
 # Generate bundle manifests and metadata, then validate generated files.
-bundle: manifests
+bundle: manifests kustomize
```
abaiken, line 245: can you add this to the make help with the optional/required args?

maskarb: added

abaiken (approved):
LGTM!!

PR #56, opened 2020-12-21
what is going on with codecov

boaz0:


crhchttp/http_cloud_dot_redhat.go
```diff
@@ -69,6 +69,7 @@ func GetMultiPartBodyAndHeaders(filename string) (*bytes.Buffer, string, error)
 	h := make(textproto.MIMEHeader)
 	h.Set("Content-Disposition", fmt.Sprintf(`form-data; name=%q; filename=%q`, "file", filename))
 	h.Set("Content-Type", "application/vnd.redhat.hccm.tar+tgz")
+	h.Set("Via", "Koku")
```
kholdaway, line 72: @boaz0 I see you set a new value.  The content-type identifies koku already.  Is this required?

kholdaway (changes requested):

crhchttp/http_cloud_dot_redhat.go
boaz0: No, I am just playing around to see why the codecov is giving me odd coverage results. Please ignore any changes from this PR. (I am not sure if it runs the codecov results if it is a draft so that's why I left this as pull request).

boaz0:
I have no idea what's wrong with you code coverage!?
When there are no tests in http_cloud_dot_redhat you're OK but when I am adding tests you're upset! Why?! :sob:

boaz0:
OK, closing this as it was resolved!

PR #62, opened 2021-01-10
COST-547 - introduce configurable PVC

maskarb:
- PVC is configurable thru the CR
- If no PVC config is provided, a default is used
- reports directory is initialized in deployment as EmptyDir
- EmptyDir is replaced with PVC
- The Pod reboots after deployment modification, and new Pod access the new PVC

Testing:
1. deploy with the default CR and ensure that the built-in default PVC is created and mounted.
2. keep the PVC from previous step, and redeploy the operator. Create a new CR and ensure that the deployment is mounted onto the previously created PVC (if reports were created in step 1, you could check that they still exist - this test is equivalent to what could happen during an operator upgrade)
3. re-deploy, create a new CR with a defined PVC. Ensure that the specked PVC is created and mounted.
4. create a new PVC. re-deploy. Create a CR with the created PVC spec. Ensure that deployment is mounted onto that created PVC
5. continuing with previous deployment, remove the PVC definition from the CR. Ensure the deployment is re-mounted onto the built-in default PVC. Also check to make sure that the PVC that was unmounted remains within the project


controllers/suite_test.go
```diff
@@ -241,24 +333,38 @@ func deleteClusterVersion() {
...
-func clusterPrep() {
+func createDeployment(ctx context.Context, deployment *appsv1.Deployment) {
+	Expect(k8sClient.Create(ctx, deployment)).Should(Succeed())
+}
+
+func deleteDeployment(ctx context.Context, deployment *appsv1.Deployment) {
+	Expect(k8sClient.Delete(ctx, deployment)).Should(Succeed())
+}
+
+// func createPVC(ctx context.Context, pvc *corev1.PersistentVolumeClaim) {
```
boaz0, line 344: should we keep it or remove this comment?

storage/storage.go
```diff
@@ -0,0 +1,172 @@
...
+	dep.Spec.Template.Spec.Volumes[s.vol.index] = *s.vol.volume
+
+	if err := s.Client.Patch(ctx, dep, patch); err != nil {
+		return false, fmt.Errorf("failed to Patch deployment: %v", err)
+	}
+	return true, nil
+}
+
+// ConvertVolume converts the EmptyDir volume in deployment to PVC
+func ConvertVolume(s *Storage, kmCfg *kokumetricscfgv1alpha1.KokuMetricsConfig) (bool, error) {
```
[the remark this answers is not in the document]
maskarb: Done!

controllers/suite_test.go
maskarb: Yes, I removed these comments.

boaz0 (approved):

abaiken (approved):
Great work on this Michael! Went through all of the test scenarios - LGTM 👍

PR #63, opened 2021-01-10
Use mounted secret path to get token and certs for prometheus connection

maskarb:
This PR simplifies certificate and token retrieval for prometheus connection.

When the manager container is created, it is mounted on the Service Account secret file path. This gives us direct access to the `service-ca.crt` and `token` files. Instead of using the `Client` to get and iterate through all of the secrets in the namespace, we directly get the files and use them in the configuration to connect to prometheus.


boaz0 (approved):

abaiken (dismissed):
We should document the process of moving the service-ca.crt and token to the expected directory for local dev but LGTM!

abaiken (approved):
Retested both locally and deployed 👍

PR #64, opened 2021-01-15
modify CSV if deployment is owned by CSV

maskarb:
Deployments owned by the CSV were not mounting onto the PVC. This PR fixes that.


Makefile
```diff
@@ -80,11 +80,8 @@ help:
 all: manager
 
 # Run tests
-ENVTEST_ASSETS_DIR=$(shell pwd)/testbin
 test: generate fmt vet manifests
-	mkdir -p ${ENVTEST_ASSETS_DIR}
```
boaz0, line 85: Why are we removing these? I thought it is essential to test Go operators?

boaz0 (approved):
All in all LGTM

abaiken (approved):
I tested the new bundle with an upgrade from 0.9.1 as well as deploying for development. Everything worked. Good work Michael!

Makefile
maskarb: These are not necessarily required. These assets were actually causing test failures, so I removed them. So long as Kubebuilder is properly installed, the tests will still run. If it turns out that we do need these, then they can be put back. But in the meantime, I think they can be removed.

PR #65, opened 2021-01-19
COST-547 -  split packaging and upload in controller

maskarb:
- packaged tar files use a timestamp in filename instead of uuid
- packaging is separated from upload in reconciler
- add packaging fields to CR (last successful packaging and list of files in storage)
- move get upload dir files to dirconfig package


controllers/kokumetricsconfig_controller.go
```diff
@@ -274,15 +274,15 @@ func GetAuthSecret(r *KokuMetricsConfigReconciler, kmCfg *kokumetricscfgv1alpha1
 func checkCycle(logger logr.Logger, cycle int64, lastExecution metav1.Time, action string) bool {
 	log := logger.WithValues("KokuMetricsConfig", "checkCycle")
 	if lastExecution.IsZero() {
-		log.Info(fmt.Sprintf("there have been no prior successful %ss to cloud.redhat.com", action))
+		log.Info(fmt.Sprintf("there have been no prior successful %ss", action))
```
abaiken, line 277: instead of removing "cloud.redhat.com" we could replace it with the api_url from the CR

controllers/kokumetricsconfig_controller.go
```diff
@@ -274,15 +274,15 @@ func GetAuthSecret(r *KokuMetricsConfigReconciler, kmCfg *kokumetricscfgv1alpha1
...
 		return true
 	}
 
 	duration := time.Since(lastExecution.Time.UTC())
 	minutes := int64(duration.Minutes())
 	log.Info(fmt.Sprintf("it has been %d minute(s) since the last successful %s", minutes, action))
 	if minutes >= cycle {
-		log.Info(fmt.Sprintf("executing %s to cloud.redhat.com", action))
+		log.Info(fmt.Sprintf("executing %s", action))
```
abaiken, line 285: same comment as above.

abaiken (commented):
LGTM! Tested. Left a comment

abaiken (approved):

controllers/kokumetricsconfig_controller.go
maskarb: Since the packaging task now uses this check cycle separately, the log would have said `there have been no prior successful packagings to cloud.redhat.com` which wouldn't make sense since the packages remain in the cluster. Removed the url so that the log makes sense. If we really need to know where the upload is going, we can view the url in the CR.

PR #66, opened 2021-01-19
COST-547 - max report count

maskarb:
The PR accomplishes several things:

- unnest PVC from Status and Status now reflects the actual PVC installed (before, it only reflected the template used, not the actual state of the system)
- confirms that all of the necessary directories exist before reconciliation (before, we were only checking for the parent dir, now we look for paret, reports, staging, and upload)
- Add new Spec field: MaxReports -> this is the maximum number of reports to keep
- Add new Status: includes the new MaxReports, the current number in storage, and the full file-path for each report in storage

The max report count does _not_ count split packages individually. If one package is split into multiple files, all of these count as 1 report.


abaiken (approved):
LGTM!

PR #70, opened 2021-02-09
COST-991 - add v1beta1 crd

maskarb:
Converting to v1beta1 will require 2 operator versions. This PR adds the v1beta1 api, converts the stored API version to v1beta1, and removes v1alpha1 from storage.

This PR also allows us to install the operator in any namespace.

The next operator version needs to update the `PROJECT` file to remove v1alpha1. The `// +kubebuilder:unservedversion` flag needs to be added to v1alpha1. During bundle creation, the v1alpha1 needs to be removed from the `owned: customerResourceDefinitions` (not sure why the bundle is adding it in).


api/v1beta1/kokumetricsconfig_types.go
```diff
@@ -0,0 +1,476 @@
...
+	// AuthType is a field of KokuMetricsConfig to represent the authentication type to be used basic or token.
+	// Valid values are:
+	// - "basic" : Enables authentication using user and password from authentication secret.
+	// - "token" (default): Uses cluster token for authentication.
+	// +kubebuilder:default="token"
+	AuthType AuthenticationType `json:"type"`
+
+	// AuthenticationSecretName is a field of KokuMetricsConfig to represent the secret with the user and password used for uploads.
+	// +optional
+	AuthenticationSecretName string `json:"secret_name,omitempty"`
```
abaiken, line 125: Previously we discussed transitioning the underscores `secret_name` to camel case `secretName` when we went to a new CRD version. Do we still want to do that?

abaiken (approved):

PR #73, opened 2021-02-10
COST-1008 - add proxy config to http transport

maskarb:
Implements an HTTP transport that uses a Proxy. The transport grabs the proxy info from the environment.

OLM should handle cluster-wide Proxy for us by adding `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` to the Pod environment. See [here](https://docs.openshift.com/container-platform/4.6/operators/admin/olm-configuring-proxy-support.html#olm-overriding-proxy-settings_olm-configuring-proxy-support).

### DIFF
--- a/crhchttp/http_cloud_dot_redhat.go
+++ b/crhchttp/http_cloud_dot_redhat.go
@@ -120,6 +120,7 @@ func SetupRequest(authConfig *AuthConfig, contentType, method, uri string, body 
 // GetClient Return client with certificate handling based on configuration
 func GetClient(authConfig *AuthConfig) HTTPClient {
 	log := authConfig.Log.WithValues("kokumetricsconfig", "GetClient")
+	transport := &http.Transport{Proxy: http.ProxyFromEnvironment}
 	if authConfig.ValidateCert {
 		// create the client specifying the ca cert file for transport
 		caCert, err := ioutil.ReadFile(cacerts)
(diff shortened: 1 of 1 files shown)


chambridge:
Not suggesting this is wrong, but the insights-operator code is a bit more in depth (could just be based on something older 🤷 ): https://github.com/openshift/insights-operator/blob/0ab6950a140c77b771b10fa7716c1dd83a03804a/pkg/insights/insightsclient/insightsclient.go#L97-L120

maskarb:
That's a good point Chris. I think our issue is actually how we are creating the transport for the client with certificates.

I updated this PR to copy in the default transport from the http package, and add the certs to that transport instead. This way our cert based client has all of these other parameters that we were ignoring previously.

crhchttp/http_cloud_dot_redhat.go
```diff
@@ -129,15 +130,10 @@ func GetClient(authConfig *AuthConfig) HTTPClient {
...
 		caCertPool.AppendCertsFromPEM(caCert)
 
-		transport := &http.Transport{TLSClientConfig: &tls.Config{RootCAs: caCertPool}}
-		return &http.Client{
-			Timeout:   30 * time.Second,
-			Transport: transport,
-		}
+		transport.TLSClientConfig = &tls.Config{RootCAs: caCertPool}
 	}
-	log.Info("configured to not using the certificate for this request")
```
abaiken, line 138: why remove this log?

maskarb: 1. this function now only has 1 return statement. If we have `validate_cert: true`, we would still log this message.
2. not sure how useful this message is anyway. We can see in the CR if `validate_cert: false`, so we are not really losing anything here.

abaiken (approved):
Nice work getting this fixed! Tested in my lab (no proxy) and verified logs in the cluster from peter with a proxy.

PR #74, opened 2021-02-11
COST-1017 - verify basic c.rh.c credentials before use

maskarb:
1. allow Secret to contain mixed-case username/password keys.
2. verify basic-auth credentials prior to source check and upload. Skip source check and upload if creds are invalid, but continue gathering reports and trimming stored reports.
3. re-attempt creds verification only if the user has changed the Secret.

Testing:
---
1. deploy operator. use basic auth -> create secret with invalid credentials
2. verify that the operator validates the credentials. verify that the operator does not perform a sources check or upload. verify that the operator does not loop immediately after determining that creds are invalid. verify that during the next reconciliation, the operator does _not_ reverify.
3. correct the Secret
4. verify that the operator successfully checks for a source and uploads reports. verify that during the next reconciliation, the operator does _not_ reverify.
 
---
1. Create a Secret with mixed-case `UseRnAME/PasSwOrD` keys.
2. Verify that the operator can still use this Secret without failing.


abaiken (approved):
Tested - LGTM! Nice job getting this fixed.

PR #76, opened 2021-02-15
do not check auth when toggle-upload is false

maskarb:
1. after first upload failure, do not try the rest
2. if upload_toggle is false, do not set or check authentication, do not check sources, do not upload files


abaiken (approved):
Tested - LGTM!

PR #80, opened 2021-02-18
remove v1alpha1 api

maskarb:
We're moving to v1beta1 api, so removing v1alpha1 altogether.


abaiken (commented):
This looks good so far - the readme/docs/makefile still refers to the v1alpha1 CR (config/samples/koku-metrics-cfg_v1alpha1_kokumetricsconfig.yaml) which is being removed in this PR so we should update those to use the beta CR.

abaiken (approved):

PR #84, opened 2021-03-10
Downstream changes with sed replacing koku/ costmanagement

abaiken:
### TESTING
* you should be able to run make upstream and see that no files are changed
* you can test the scratch build that is based off this commit by either running a scratch build yourself from the downstream repo or using the make command with my username (since i have already mirrored the image in quay):
```
make deploy-user-scratch USER=abaiken
```
You can make sure that the operator commit in the status matches the commit on this branch **NOTE:** It might not be the latest commit bc I tweaked the makefile


config/manager/kustomization.yaml
```diff
@@ -1,2 +1,8 @@
 resources:
 - manager.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
```
maskarb, line 3: We should probably exclude this file from git.

config/samples/kustomization.yaml
```diff
@@ -1,3 +1,3 @@
 ## This file is auto-generated, do not modify ##
 resources:
-- koku-metrics-cfg_v1beta1_kokumetricsconfig.yaml
+- costmanagement-metrics-cfg_v1beta1_costmanagementmetricsconfig.yaml
```
maskarb, line 3: does this file exist?

docs/local-development.md
```diff
@@ -3,7 +3,7 @@
 ## Pre-reqs
 
 * Access to a 4.3+ Openshift cluster
-* A clone of [korekuta-go-operator](https://github.com/project-koku/koku-metrics-operator)
+* A clone of [korekuta-go-operator](https://github.com/project-costmanagement/costmanagement-metrics-operator)
```
maskarb, line 6: We should update `korekuta-go-operator`. Didn't realize we never changed this reference.

maskarb: But on a different note, should we just exclude the docs folder in this branch? I'd be ok with adding this to the `remove files` in the Makefile.

myersCody (dismissed):
LGTM! 
Awesome work ash! Cachito is really cool, I wish we had that option back in the Discovery days.

docs/local-development.md
abaiken: Added to the remove files

config/samples/kustomization.yaml
abaiken: It does now - I added copying it to the correct name as part of the `make upstream` command

config/manager/kustomization.yaml
abaiken: Fixed - I removed my changes. It is in the gitignore but this file has always been odd

maskarb (approved):

PR #89, opened 2021-04-01
update source creation steps

maskarb:
1. make the source creation docs a little more generic. The sources wizard should be the real guide, not the docs here.
2. added a `faqs` doc. Includes steps for forcing v0.9.4 -> v0.9.5 upgrade.


abaiken (approved):
Looks good!

PR #100, opened 2021-07-15
COST-1551: OCP logging & status reflection

abaiken:
Add logs to the file upload


packaging/packaging.go
```diff
@@ -175,14 +183,28 @@ func (p *FilePackager) writeTarball(tarFileName, manifestFileName string, archiv
...
+	files = append(files, manifestFileName)
 	for idx, filePath := range archiveFiles {
 		if strings.HasSuffix(filePath, ".csv") {
 			uploadName := p.uid + "_openshift_usage_report." + strconv.Itoa(idx) + ".csv"
+			files = append(files, uploadName)
 			if err := p.addFileToTarWriter(uploadName, filePath, tw); err != nil {
 				return fmt.Errorf("writeTarball: failed to create tar file: %v", err)
 			}
+		} else {
+			files = append(files, filePath)
```
maskarb, line 196: Why is this `else` here?

controllers/kokumetricsconfig_controller.go
```diff
@@ -470,8 +470,13 @@ func uploadFiles(r *KokuMetricsConfigReconciler, authConfig *crhchttp.AuthConfig
 			return err
 		}
 		ingressURL := kmCfg.Status.APIURL + kmCfg.Status.Upload.IngressAPIPath
-		uploadStatus, uploadTime, err := crhchttp.Upload(authConfig, contentType, "POST", ingressURL, body)
+		uploadInfo, exists := fileInfo[file]
```
maskarb, line 473:
```suggestion
		uploadInfo, exists := fileInfo[file]
		if !exists {
			<log something>
			continue // skip processing this file
		}
```

This `exists` should be immediately checked. If the file isn't there, then I'm pretty sure `uploadInfo` will be `nil` and the logic below will fail.

controllers/kokumetricsconfig_controller.go
```diff
@@ -481,6 +486,9 @@ func uploadFiles(r *KokuMetricsConfigReconciler, authConfig *crhchttp.AuthConfig
 		if strings.Contains(uploadStatus, "202") {
 			kmCfg.Status.Upload.LastSuccessfulUploadTime = uploadTime
 			// remove the tar.gz after a successful upload
+			if exists {
```
maskarb, line 489: If the `exists` is checked above, then this `if` statement can be removed.

packaging/packaging.go
abaiken: It is pre-emptive because i am going to add a file to the payload in a future PR with the CR status/any errors and it will be in json format

maskarb: 🤔  Why would we add a new file instead of just adding it into a field in the manifest file?

abaiken: I'll remove it for now, but the story is left undefined on if we want to add to the manifest or add a new file so I might add it back when I pick up that issue (just depends on the what makes the most sense)

controllers/kokumetricsconfig_controller.go
abaiken: Added a check for exists immediately below grabbing the file info

packaging/packaging.go
```diff
@@ -498,3 +500,45 @@ func (p *FilePackager) PackageReports() error {
...
+		if err != nil {
+			log.Info("Could not read from the buffer.")
+			return fileInfo, err
+		}
+		s := buff.String()
+		if err := json.Unmarshal([]byte(s), &fileInfo); err != nil {
+			log.Info("Could not load the manifest json.")
+			return fileInfo, err
+		}
+	}
```
maskarb, line 542:
```suggestion
		}
	break // exit `for` loop after processing the manifest.json
	}
```

packaging/packaging.go
```diff
@@ -498,3 +500,46 @@ func (p *FilePackager) PackageReports() error {
...
+		if err != nil {
+			log.Info("Could not read from the buffer.")
+			return fileInfo, err
+		}
+		s := buff.String()
+		if err := json.Unmarshal([]byte(s), &fileInfo); err != nil {
+			log.Info("Could not load the manifest json.")
+			return fileInfo, err
+		}
+	break // exit `for` loop after processing the manifest.json
```
maskarb, line 542:
```suggestion
		break // exit `for` loop after processing the manifest.json
```

abaiken: fixed

maskarb (approved):
looks good! thanks for making all these changes!

PR #103, opened 2021-07-27
Add a statement to check if either the cluster version or cluster id isn't there

abaiken:
caught this during release testing


maskarb (approved):
lgtm!

PR #109, opened 2021-11-26
COST-2096: Update prom queries to exclude non-running containers

chambridge:
* Update request and limit Prometheus queries for cpu & memory

Associated with: https://issues.redhat.com/browse/COST-2096

### DIFF
--- a/collector/queries.go
+++ b/collector/queries.go
@@ -127,7 +127,7 @@ var (
 	podQueries = &querys{
 		query{
 			Name:        "pod-limit-cpu-cores",
-			QueryString: "sum(kube_pod_container_resource_limits{resource='cpu'}) by (pod, namespace, node)",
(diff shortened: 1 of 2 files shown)


ddonahue007 (approved):
LGTM

collector/queries.go
```diff
@@ -127,7 +127,7 @@ var (
 	podQueries = &querys{
 		query{
 			Name:        "pod-limit-cpu-cores",
-			QueryString: "sum(kube_pod_container_resource_limits{resource='cpu'}) by (pod, namespace, node)",
+			QueryString: "sum(kube_pod_container_resource_limits{resource='cpu'} * on(container, pod, namespace) group_right kube_pod_container_status_ready)  by (pod, namespace, node)",
```
maskarb, line 130: These should use `group_left`, otherwise we lose the `node` in the result.

maskarb (changes requested):
This looks good, but we need one change.

PR #110, opened 2021-12-06
add exponential backoff for controller errors

maskarb:
This PR does 2 things:
- add exponential backoff for reconciliation errors. 
- Return errors from prometheus report generation errors.


maskarb:
This was implemented in https://github.com/project-koku/koku-metrics-operator/pull/210

PR #111, opened 2021-12-06
break long running queries into smaller chunks

maskarb:
This PR breaks longer running queries into 5 minute chunks. The results from these chunks are aggregated together into a new `model.Matrix` and then the rest of the logic is performed on this new matrix.


abaiken (commented):
Code looks good to me & I downloaded reports and looked at them and don't see any red flags. I think we should upload a report and make sure it is processed correctly before merging this just to be safe.

abaiken (approved):

PR #126, opened 2022-01-25
Increase context timeout to 90 seconds

maskarb:
COST-2127

* make context timeout configurable
* default to 90 seconds


abaiken (dismissed):

abaiken (approved):

collector/prometheus_test.go
```diff
@@ -21,6 +21,7 @@ import (
 )
 
 var trueDef = true
+var defaultContextTimeout int64 = 90
```
abaiken, line 24: shouldn't we set this to 120

PR #135, opened 2022-05-06
[COST-992] filter out empty volumename to prevent many to many issue on PVCs

cgoodfred:
This issue was replicated and the many to many for PVC's can come from an empty volume name. Filtering out null volumenames in the prometheus query should address that.


maskarb (approved):
lgtm!

PR #139, opened 2022-05-23
create MetricsConfig to inherit CR

maskarb:
Today, releasing downstream is difficult because of the conversion from `KokuMetrics` to `CostmanagementMetrics`. We originally designed things to pass the `KokuMetricsConfig` type around many places throughout the code base. This PR creates a generic `MetricsConfig` which will inherit everything from either `KokuMetricsConfig` or `CostmanagementMetricsConfig`. This enables us to define functions that require the `MetricsConfig` instead of the specific Koku/Costmanamgent config. For future downstream releases, we won't have to convert the name `koku` to `cost-management` in as many places which should make our downstream releases and code reviews much simpler.


maskarb:
I just cannot get this to work correctly.

PR #140, opened 2022-06-27
[COST-2285] - fix many-to-many query error

maskarb:
* switch the `by (pod, namespace, node)` to `without (instance, uid)` which will prevent `many-to-many matching not allowed` errors caused by the `instance` and `uid` fields.
* add `{namespace,node,pod}!=''` because these fields are required for proper data aggregation
* add `MetricKey` to pod labels to prevent line items that are missing pod/namespace


abaiken (approved):
LGTM

PR #158, opened 2023-01-26
[COST-3422] - gather all available data in the current month on install

maskarb:
- some reorganization and renaming of code
- gather all available data in current month on CR creation
- Add API field to disable previous data collection


collector/prometheus.go
```diff
@@ -65,113 +58,140 @@ func getBearerToken(tokenFile string) (config.Secret, error) {
...
 	}
-	promCfg := &PrometheusConfig{
-		Address: kmCfg.SvcAddress,
-		CAFile:  filepath.Join(serviceaccountPath, certKey),
-		SkipTLS: *kmCfg.SkipTLSVerification,
+}
+
+type PrometheusConfigurationSetter func(ps *kokumetricscfgv1beta1.PrometheusSpec, c *PrometheusCollector) error
+
+func SetPrometheusConfig(ps *kokumetricscfgv1beta1.PrometheusSpec, c *PrometheusCollector) error {
```
samdoran, line 84: Short variable names hinder readability. Even though this function is pretty short, it would still be better to have more descriptive variable and parameter names.
```suggestion
func SetPrometheusConfig(promspec *kokumetricscfgv1beta1.PrometheusSpec, collector *PrometheusCollector) error {
```

maskarb: I understand that perspective but `c` for `PrometheusCollector` is and has been used everywhere else. `ps` was `kmCfg` which made no sense. If there really is an issue with readability, we can address that in a follow up PR.

controllers/kokumetricsconfig_controller.go
```diff
@@ -674,7 +689,21 @@ func (r *KokuMetricsConfigReconciler) Reconcile(ctx context.Context, req ctrl.Re
 	}
 
 	// attempt to collect prometheus stats and create reports
-	collectPromStats(r, kmCfg, dirCfg)
+	if err := getPromCollector(r, kmCfg); err != nil {
+		log.Error(err, "failed to get prometheus connection")
+		return ctrl.Result{RequeueAfter: time.Minute * 2}, err // give things a break and try again in 2 minutes
```
samdoran, line 694: You could use exponential backoff here starting with a smaller value.

controllers/kokumetricsconfig_controller.go
```diff
@@ -517,33 +513,43 @@ func uploadFiles(r *KokuMetricsConfigReconciler, authConfig *crhchttp.AuthConfig
 	return nil
 }
 
-func collectPromStats(r *KokuMetricsConfigReconciler, kmCfg *kokumetricscfgv1beta1.KokuMetricsConfig, dirCfg *dirconfig.DirectoryConfig) {
-	log := log.WithName("collectPromStats")
+func getTimeRange(r *KokuMetricsConfigReconciler, kmCfg *kokumetricscfgv1beta1.KokuMetricsConfig) (time.Time, time.Time) {
+	start := time.Now().UTC().Truncate(time.Hour).Add(-time.Hour) // start of previous full hour
+	end := start.Add(59*time.Minute + 59*time.Second)
+	if kmCfg.Status.Prometheus.LastQuerySuccessTime.IsZero() && r.InCluster && !r.disablePreviousDataCollection {
+		start = time.Date(start.Year(), start.Month(), 1, 0, 0, 0, 0, start.Location())
```
samdoran, line 520: So this will at most gather data back to beginning of the current month?

Would it be better to gather a fixed number of days of data?

What happens if we request data for a date range where there is no data? I believe that scenario is handled in `collectPromStats` just wanted to make sure.

collector/prometheus.go
samdoran: Sounds good.

controllers/kokumetricsconfig_controller.go
maskarb: Why?

controllers/kokumetricsconfig_controller.go
maskarb: The best solution is to gather whatever is available, but we don't know when the data starts. Given that most billing starts at the beginning of a month, that makes it a good place to start.

If there is no data, we just move to the next hour.

controllers/kokumetricsconfig_controller.go
samdoran: To potentially reduce overall wait time and/or avoid overwhelming the service.

It may not make sense in this particular scenario, but it's a good technique with decentralized clients interacting with a central service.

Not a blocker, just a comment. It's something that always comes to mind when I see fixed wait times.

controllers/kokumetricsconfig_controller.go
samdoran: That makes sense. I hadn't thought about lining up with billing date. I was just thinking "get all the data available".

chambridge (approved):
LGTM

PR #159, opened 2023-01-29
update dependencies and Go version

maskarb:
- update dependencies
- update to Go 1.18
- add loggers at the package level instead of passing a logger thru different structs
- add outline for a PrometheusRule (currently commented out and unused)


controllers/prometheus.go
```diff
@@ -0,0 +1,64 @@
+package controllers
+
+// import (
```
chambridge, line 3: any reason this file was all just commented out?

Should it just be deleted if it's no longer used?

chambridge (commented):
minor comment about `prometheus.go` file

controllers/prometheus.go
maskarb: I added that one in this PR. I plan to use it when adding the new queries. Lint wouldn't let me add those functions since technically they are unused at the moment, so I just commented out the whole file.

chambridge (approved):

PR #162, opened 2023-02-16
if reconciliation is successful, requeue until next hour

maskarb:


maskarb:
After some thinking, this would cause a spike on the hour when reports are uploaded. At least with the way things work now, uploads are likely randomly distributed throughout the hour during those every-5-min reconciliations.

PR #165, opened 2023-02-17
variable reorg and simplification

maskarb:
This PR builds on https://github.com/project-koku/koku-metrics-operator/pull/161. Move the prefix `koku` to the API directory so that it can be imported elsewhere. When we need to change `koku` to `costmanagement`, we will only need to update one place instead of many.


myersCody (approved):
LGTM

PR #167, opened 2023-02-20
[COST-3446] - update Resource Optimization metrics

maskarb:
* remove `deployment_name` and replace with `owner_kind`, `owner_name`, `workload`, and `workload_type`.


samdoran (approved):

myersCody (approved):
LGTM

PR #169, opened 2023-03-02
[COST-3422] Use retention period

maskarb:
* set the data collection period on CR creation to either 14d or the value configured in the monitoring stack. OR a max of 90d.


chambridge (approved):
nice

PR #171, opened 2023-03-13
read ConfigMap directly from api-server

maskarb:
* use APIReader to get the openshift-monitoring configMap directly from the api-server instead of from cache. Since the operator is namespaced, it is not easy or straightforward to add other namespaced resource (e.g. ConfigMaps) to the cache. Since we only read this configMap once during the lifetime of the CR, we do not need to worry about making too many requests against the api-server
* Set the start and end dates in the manifest from the ROS report if the Pod report does not exist.


myersCody (approved):
LGTM! Thanks again for the explanation through slack. 🙇

PR #173, opened 2023-03-17
remove required field from CR status

maskarb:
* Add `omit-empty` to the `previous_data_collected` status field so upgrades from old versions of the operator do not fail.


api/v1beta1/zz_generated.deepcopy.go
```diff
@@ -1,7 +1,7 @@
 // +build !ignore_autogenerated
 
 //
-// Copyright 2021 Red Hat Inc.
+// Copyright 2023 Red Hat Inc.
```
samdoran, line 4: The copyright year should reflect the year content was first created.

Makefile
```diff
@@ -120,6 +120,7 @@ uninstall: manifests kustomize
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
 deploy: manifests kustomize
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMAGE_SHA}
+	# $(KUSTOMIZE) build config/default > out.yaml
```
samdoran, line 123: Do we need to keep this line?

samdoran (dismissed):

Makefile
maskarb: No, but I've been using it during dev. But I'll remove the Makefile from this PR.

samdoran (approved):

PR #185, opened 2023-05-02
update CI workflow

maskarb:


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,98 @@
+name: CI
+
+on:
+  workflow_dispatch:
```
samdoran, line 4: Is this for manually running the workflow?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,98 @@
...
+        uses: actions/download-artifact@v3
+        with:
+          name: coverage
+
+      - name: Upload coverage
+        uses: codecov/codecov-action@v3.1.1
+        with:
+          file: ./cover.out
+          flags: unittests
+          token: ${{ secrets.CODECOV_TOKEN }}
```
samdoran, line 97: The token should only be needed if the repo is private.

.github/workflows/ci.yaml
maskarb: yes.

PR #188, opened 2023-05-19
add more logic to prevent resetting the packaging time

maskarb:
* don't reset the LastSuccessfulPackagingTime as a mechanism for forcing packaging
  * separate out the bit that actually does the packaging into `func forcePackageFiles` and call this directly
  * the original `packageFiles` func calls `forcePackageFiles` after it's done the cycle check
* use a new variable `newInstall` to determine if we need to force a packaging after an upgraded operator

### DIFF
--- a/controllers/kokumetricsconfig_controller.go
+++ b/controllers/kokumetricsconfig_controller.go
@@ -691,7 +691,7 @@ func (r *MetricsConfigReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 	r.initialDataCollection = false
 	packager.FilesAction = packaging.CopyFiles
-	if endTime.Hour() == HOURS_IN_DAY {
+	if endTime.Hour() == HOURS_IN_DAY && cr.Status.Packaging.LastSuccessfulPackagingTime.Hour() != HOURS_IN_DAY {
 		// when we've reached the end of the day, force packaging to occur to generate the daily report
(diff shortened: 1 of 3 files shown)


controllers/kokumetricsconfig_controller.go
```diff
@@ -436,6 +436,10 @@ func packageFiles(p *packaging.FilePackager, cr *metricscfgv1beta1.MetricsConfig
 		return
 	}
 
+	forcePackageFiles(p, cr)
+}
+
+func forcePackageFiles(p *packaging.FilePackager, cr *metricscfgv1beta1.MetricsConfig) {
```
samdoran, line 442: This is great separation. Distinct functions that do specific things are much better than passing a toggle that changes behavior.

Not a blocker, just a quibble over the name: this new function could be `packageFiles` since that is exactly what it does. The original could be `packageFilesDuringCycle` or something that better communicates that it only packages files when needed.

I think `maybePackageFiles` would be accurate but probably not the best name. 😆

samdoran (dismissed):

adberglund (approved):

controllers/kokumetricsconfig_controller.go
maskarb: I can get behind changing the names. Lemme push up a commit real quick.

samdoran (approved):

PR #189, opened 2023-05-21
add String to FilesAction for clearer logging

maskarb:
* convert `FilesAction` to a struct so we can define a `String` method
  * the struct contains a single field `action` which is the previous `func` definition of the `FilesAction` type
  * The `String` method can be used in logging so we know which action is being performed (`copy` or `move`)


maskarb:
This doesn't work the way I intended. I am not sure it is possible to have a string representation of a function type.

PR #195, opened 2023-08-03
[COST-3758] retry failed operator queries up to 5 times

maskarb:
* retry a given failed query up to 5 times
* retry an entire time range up to 5 times
  * track a count of how many times a particular start time has been tried. If we've tried a range 5 times, we skip it
  * we will maintain this tracking until we've had a successful query. If a success only comes after we've upgraded an operator, we will retry as far back as possible based on the retention period and the lastquerysuccesstime.
  * If a success comes without updating the operator, then the missed data is not collected.
* update the txt_replace.py script to use argparse and accept an optional namespace argument so you can generate a CSV with that namespace so you can deploy the CSV without manually updating the namespace

With the changes to the script, I've been using this series of make commands in order to deploy the operator to a cluster:
```
make docker-build-no-test IMG=quay.io/$USERNAME/koku-metrics-operator:v$VERSION; make docker-push IMG=quay.io/$USERNAME/koku-metrics-operator:v$VERSION; make bundle IMG=quay.io/$USERNAME/koku-metrics-operator:v$VERSION NAMESPACE=koku-metrics-operator; oc apply -f koku-metrics-operator/2.0.1/manifests/koku-metrics-operator.clusterserviceversion.yaml
```

In order to deploy an operator via a CSV like this, an OperatorGroup needs to be created:
```
apiVersion: v1
kind: Namespace
metadata:
  labels:
    app: koku-metrics-operator
    control-plane: controller-manager
  name: koku-metrics-operator
---
apiVersion: operators.coreos.com/v1
kind: OperatorGroup
metadata:
  name: koku-metrics-operator
spec:
  targetNamespaces:
  - koku-metrics-operator
```

I've built this little "proxy" server so that we can simulate slow queries:
https://github.com/maskarb/devfile-sample-go-basic

In the developer view of an Openshift cluster, that repo can be imported and deployed. Then KokuMetricsConfig can be updated like this:
```
  prometheus_config:
    collect_previous_data: true
    context_timeout: 10
    disable_metrics_collection_cost_management: false
    disable_metrics_collection_resource_optimization: false
    service_address: 'http://prom-test-server-koku-metrics-operator.apps-crc.testing' (this route is found in the cluster UI)
```

With this service address, the prometheus traffic will flow thru the "proxy" server and will randomly sleep for longer than 10 seconds. This will cause a timeout to occur which is visible in the operator logs. Once that query timeout occurs, that individual query will be requeued and tried again.

### DIFF
--- a/collector/prometheus.go
+++ b/collector/prometheus.go
@@ -239,7 +239,8 @@ func (c *PrometheusCollector) getQueryRangeResults(queries *querys, results *map
 
 	if len(queriesToRetry) > 0 {
 		retries--
-		waitTime := time.Duration(math.Pow(2, float64(MaxRetries-retries))) * time.Second
+		sleep := math.Max(math.Pow(2, float64(MaxRetries-retries)), 1)
+		waitTime := time.Duration(sleep) * time.Second
 		log.Info(fmt.Sprintf("retrying failed queries after %s seconds", waitTime))
 		time.Sleep(waitTime)
 		return c.getQueryRangeResults(&queriesToRetry, results, retries)
@@ -279,7 +280,8 @@ func (c *PrometheusCollector) getQueryResults(ts time.Time, queries *querys, res
 
 	if len(queriesToRetry) > 0 {
 		retries--
-		waitTime := time.Duration(math.Pow(2, float64(MaxRetries-retries))) * time.Second
+		sleep := math.Max(math.Pow(2, float64(MaxRetries-retries)), 1)
+		waitTime := time.Duration(sleep) * time.Second
 		log.Info(fmt.Sprintf("retrying failed queries after %s seconds", waitTime))
 		time.Sleep(waitTime)
 		return c.getQueryResults(ts, &queriesToRetry, results, retries)
--- a/collector/prometheus.go
+++ b/collector/prometheus.go
@@ -244,8 +244,10 @@ func (c *PrometheusCollector) getQueryRangeResults(queries *querys, results *map
 	return nil
 }
 
-func (c *PrometheusCollector) getQueryResults(ts time.Time, queries *querys, results *mappedResults) error {
+func (c *PrometheusCollector) getQueryResults(ts time.Time, queries *querys, results *mappedResults, retries int) error {
 	log := log.WithName("getQueryResults")
+
+	queriesToRetry := querys{}
 
 	for _, query := range *queries {
 		ctx, cancel := context.WithTimeout(context.Background(), c.ContextTimeout)
@@ -253,7 +255,13 @@ func (c *PrometheusCollector) getQueryResults(ts time.Time, queries *querys, res
 
 		queryResult, warnings, err := c.PromConn.Query(ctx, query.QueryString, ts)
 		if err != nil {
-			return fmt.Errorf("query: %s: error querying prometheus: %v", query.QueryString, err)
+			if retries > 0 {
+				log.Info(fmt.Sprintf("query `%s` failed, appending to queries to retry", query.Name))
+				queriesToRetry = append(queriesToRetry, query)
+				continue
+			} else {
+				return fmt.Errorf("query: %s: error querying prometheus: %v", query.QueryString, err)
+			}
 		}
 		if len(warnings) > 0 {
 			log.Info("query warnings", "Warnings", warnings)
@@ -265,5 +273,14 @@ func (c *PrometheusCollector) getQueryResults(ts time.Time, queries *querys, res
 
 		results.iterateVector(vector, query)
 	}
+
+	if len(queriesToRetry) > 0 {
+		retries--
+		waitTime := time.Duration(5-retries) * time.Second
+		log.Info(fmt.Sprintf("retrying failed queries after %s seconds", waitTime))
+		time.Sleep(waitTime)
+		return c.getQueryResults(ts, &queriesToRetry, results, retries)
+	}
+
 	return nil
 }
--- a/collector/prometheus.go
+++ b/collector/prometheus.go
@@ -206,16 +206,23 @@ func (c *PrometheusCollector) GetPromConn(
 	return nil
 }
 
-func (c *PrometheusCollector) getQueryRangeResults(queries *querys, results *mappedResults) error {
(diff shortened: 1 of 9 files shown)


chambridge (approved):
LGTM

collector/prometheus.go
```diff
@@ -224,22 +231,37 @@ func (c *PrometheusCollector) getQueryRangeResults(queries *querys, results *map
 		if !ok {
 			return fmt.Errorf("expected a matrix in response to query, got a %v", queryResult.Type())
 		}
-
 		results.iterateMatrix(matrix, query)
 	}
+
+	if len(queriesToRetry) > 0 {
+		retries--
+		waitTime := time.Duration(5-retries) * time.Second
```
samdoran, line 239: You could use exponential backoff here rather than a linear algorithm. If there is potential for a thundering herd, adding some jitter to the wait time will mitigate that.

scripts/txt_replace.py
```diff
@@ -65,13 +72,17 @@ def fix_dockerfile(version_tuple):
...
+    config = vars(args)
+    print(config)
 
-    version_tuple = sys.argv[1:]
-    check_version(version_tuple)
+    check_version(config["version"], config["previous_version"])
 
-    fix_csv(version_tuple)
-    fix_dockerfile(version_tuple)
+    fix_csv(**config)
```
samdoran, line 87: Tightly coupling the arg names to the function params is convenient but brittle. This is less elegant than what you have but is more maintainable.
```suggestion
    fix_csv(args.version, args.previous_version, args.image_sha, args.namespace)
```

scripts/txt_replace.py
```diff
@@ -65,13 +72,17 @@ def fix_dockerfile(version_tuple):
...
+    fix_dockerfile(config["version"])
```
samdoran, line 88: You could just use the namespace attribute directly.
```suggestion
    fix_dockerfile(args.version)
```

scripts/txt_replace.py
```diff
@@ -65,13 +72,17 @@ def fix_dockerfile(version_tuple):
         replace(filename, k, v)
 
 if __name__ == "__main__":
-    nargs = len(sys.argv)
-    if nargs != 4:
-        print("usage: %s VERSION PREVIOUS_VERSION IMAGE_SHA" % path.basename(sys.argv[0]))
-        exit()
+    parser = argparse.ArgumentParser(description="Script for updating the appropriate fields of the CSV",
+                                 formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument("-n", "--namespace", help="namespace used for testing")
```
samdoran, line 77: This should default to `""` to avoid inserting `'namespace: None'` in `fix_csv()`.
```suggestion
    parser.add_argument("-n", "--namespace", help="namespace used for testing", default="")
```

samdoran (approved):
A handful of comments/suggestions but nothing major.

collector/prometheus.go
maskarb: How do we determine the risk of thundering herd? I can implement an exponential backoff, but because I don't understand the risk, I don't understand the benefit. I only see a downside because unless I change some testing logic, this is going to make our tests take a long time.

scripts/txt_replace.py
```diff
@@ -65,13 +72,17 @@ def fix_dockerfile(version_tuple):
         replace(filename, k, v)
 
 if __name__ == "__main__":
-    nargs = len(sys.argv)
-    if nargs != 4:
-        print("usage: %s VERSION PREVIOUS_VERSION IMAGE_SHA" % path.basename(sys.argv[0]))
-        exit()
+    parser = argparse.ArgumentParser(description="Script for updating the appropriate fields of the CSV",
+                                 formatter_class=argparse.ArgumentDefaultsHelpFormatter)
```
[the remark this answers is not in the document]
maskarb: done.

scripts/txt_replace.py
maskarb: done.

scripts/txt_replace.py
maskarb: done.

collector/prometheus.go
samdoran: > How do we determine the risk of thundering herd?

Consider the number of copies of this code running and if there are potential triggers that could cause them all to run synchronously, such as a process restart or time based run. It may not be an issue in the current architecture, but it's just something to consider when implementing retry code.

Adding some random fractional jitter to the sleep time helps break up the requests so they don't all arrive at the same time even if the processes were all triggered at the same time.

> I can implement an exponential backoff, but because I don't understand the risk, I don't understand the benefit

Exponential backoff can help with two things: 1) Reducing overall wait time and 2) Reducing the number of requests made to a resource.

If it doesn't accomplish those two things then it doesn't add any benefit over a linear retry timer.

For example, say the wait time is 0.5, then 1, then 2 and it succeeded on the third retry. That's less wait time than five seconds per retry.

maskarb: The risk here is pretty low then. Only a single operator runs on a cluster, and there is only a single CR created that the operator handles. Everything runs synchronously.

While the retry time was not exponential, it wasn't just 5 seconds between retries; it was 1, 2, 3, 4, 5 seconds.

samdoran: Sounds good. Sorry I misread the sleep times.

collector/prometheus.go
```diff
@@ -236,7 +239,7 @@ func (c *PrometheusCollector) getQueryRangeResults(queries *querys, results *map
 
 	if len(queriesToRetry) > 0 {
 		retries--
-		waitTime := time.Duration(5-retries) * time.Second
+		waitTime := time.Duration(math.Pow(2, float64(MaxRetries-retries))) * time.Second
```
samdoran, line 242: This might be an edge case, but if `retries` is greater than `MaxRetries`, the wait duration ends up being 0s until they are equal. 
```suggestion
                sleep := math.Max(math.Pow(2, float64(MaxRetries-retries), 1)
		waitTime := time.Duration(sleep) * time.Second
```

maskarb: done.

samdoran (approved):

PR #196, opened 2023-08-09
[COST-4080] Replace default source name with cluster-uuid if CR is not updated

djnakabaale:
## Jira Ticket

[COST-4080](https://issues.redhat.com/browse/COST-4080)

## Description

- Use an empty string placeholder vs `INSERT-SOURCE-NAME` in the spec source name field.
- if the spec source name value is an empty string, replace it with cluster id.


maskarb (approved):
lgtm! Nicely done!

PR #210, opened 2023-10-04
[COST-4230] Add service account auth type

djnakabaale:
## Jira Ticket

[COST-4230](https://issues.redhat.com/browse/COST-4230)

## Description

This change will add service-account authentication type

**Prerequisites:** 
- Requires an OpenShift cluster. You can request one via cluster bot in slack or the [Red Hat  OpenShift Local](https://access.redhat.com/documentation/en-us/red_hat_openshift_local/2.5/html/getting_started_guide/installation_gsg#installing_gsg) version.
- Requires service account credentials, `client_id` and `client_secret`

## Testing

1. Checkout Branch
3. Launch shell and follow the steps below:

```sh
#  log into your OpenShift cluster, 
oc login -u kubeadmin -p <PASS> <server>

# create a koku-metrics-operator project: 
oc new-project koku-metrics-operator`

# set koku-metrics-operator to be the working namespace
oc project koku-metrics-operator

export USERNAME=<your_quay_username>
export QUAY_TAG=v3.sa.test
export QUAY_IMG=quay.io/$USERNAME/koku-metrics-operator:$QUAY_TAG


# Build & push the image, and make sure the repository is public:
make docker-build IMG=$QUAY_IMG
make docker-push IMG=$QUAY_IMG

# Deploy operator
make deploy IMG=$QUAY_IMG

```

3. In the `testing` folder, create the secret and kokumetricsconfig yamls
```yml
# sa-auth-secret-template.yaml
apiVersion: v1
kind: Secret
metadata:
  name: dev-auth-secret
data:
  client_id: <base64_client_id>
  client_secret: <base64_client_secret>



# sa_koku-metrics-cfg_v1beta1_kokumetricsconfig.yml
apiVersion: koku-metrics-cfg.openshift.io/v1beta1
kind: KokuMetricsConfig
metadata:
  name: kokumetricscfg-sample
spec:
  packaging: {}
  prometheus_config:
    collect_previous_data: true
    service_address: https://thanos-querier-openshift-monitoring.apps-crc.testing
    skip_tls_verification: true
  upload:
    validate_cert: false
  source:
    check_cycle: 1440
    create_source: false
    name: ""
    sources_path: /api/sources/v1.0/
  authentication:
    type: service-account
    secret_name: dev-auth-secret
    token_url: https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token
  api_url: 'https://console.redhat.com/'


```

4. You should see a pod created and pod logs similar to this:
    <details>
    
    ```log

    2023-11-30T14:24:31.186904384Z INFO setup starting manager
    2023-11-30T14:24:31.187596859Z INFO controller-runtime.metrics Starting metrics server
    2023-11-30T14:24:31.188304348Z INFO controller-runtime.metrics Serving metrics server {"bindAddress": ":8080", "secure": false}
    I1130 14:24:31.188832 1 leaderelection.go:250] attempting to acquire leader lease koku-metrics-operator/91c624a5.openshift.io...
    I1130 14:24:52.378759 1 leaderelection.go:260] successfully acquired lease koku-metrics-operator/91c624a5.openshift.io
    ```

    </details>

5. Deploy your secret and kokumetricsconfig

```bash
# deploy dev secret:
oc apply -f testing/sa-auth-secret-template.yaml 

# deploy local crd:
oc apply -f testing/sa_koku-metrics-cfg_v1beta1_kokumetricsconfig.yaml
```

If credentials are valid, should see the operator attempting to get or create a source (depending on the `create_source` value ) as well payload uploads.

## Notes

- It's likely to see errors related to permissions if your service-account credentials are not associated with a user group.
- The operator will only attempt to create a source if `create_source: true` and no OpenShift source is registered with the same name and cluster id.

...


crhchttp/config.go
```diff
@@ -6,19 +6,96 @@
...
+	log := log.WithName("GetAccessToken")
+
+	// Prepare the POST data
+	data := neturl.Values{}
+	data.Set("client_id", ac.ServiceAccountData.ClientID)
+	data.Set("client_secret", ac.ServiceAccountData.ClientSecret)
+	data.Set("grant_type", ac.ServiceAccountData.GrantType)
+
+	// // Making the HTTP POST request
+	resp, err := http.Post(tokenURL, "application/x-www-form-urlencoded", bytes.NewBufferString(data.Encode()))
```
maskarb, line 67: After thinking about this a little bit, I think we need to use `context`, and specify a timeout (just in case).

That requires using `http.NewRequestWithContext` and a `client.Do`.

```
ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second) <<<< or even better to pass `ctx` into `GetAccessToken`, and use that here
defer cancel()
req, err := http.NewRequestWithContext(ctx, http.MethodPost, tokenURL, bytes.NewBufferString(data.Encode()))
// handle err
```
and then:
```
resp, err := http.DefaultClient.Do(req)
```
It IS a little annoying to handle yet another error with `newrequestwithcontext`, but still good to do regardless.

maskarb (commented):
Initial feedback. This is a great start! Many of these are a little nitpicky, but they'll align with idiomatic Go.

I also still need to look over the unit tests.

api/v1beta1/metricsconfig_types.go
```diff
@@ -278,6 +289,9 @@ type AuthenticationStatus struct {
 	// LastVerificationTime is a field of KokuMetricsConfig to represent the last time credentials were verified.
 	// +nullable
 	LastVerificationTime *metav1.Time `json:"last_credential_verification_time,omitempty"`
+
+	// TokenURL is a filed of KokuMetricsConfig to represent the url used to generate a service account token
```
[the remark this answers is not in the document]
djnakabaale: updated

api/v1beta1/metricsconfig_types.go
```diff
@@ -278,6 +289,9 @@ type AuthenticationStatus struct {
...
+	TokenURL string `json:"token_url"`
```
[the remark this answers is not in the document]
djnakabaale: yes 👍🏾, I've added this

controllers/kokumetricsconfig_controller.go
```diff
@@ -364,6 +429,17 @@ func validateCredentials(r *MetricsConfigReconciler, handler *sources.SourceHand
...
 
+	// Service  Account authentication check
+	if cr.Spec.Authentication.AuthType == metricscfgv1beta1.ServiceAccount {
+		err := handler.Auth.GetAccessToken(cr.Spec.Authentication.TokenURL)
+		if err != nil {
+			errorMsg := fmt.Sprintf("Invalid client credentials provided. Correct the client-id / client-secret in `%s`. Updated credentials will be re-verified during the next reconciliation.", cr.Spec.Authentication.AuthenticationSecretName)
+			log.Info(errorMsg)
+			cr.Status.Authentication.AuthErrorMessage = errorMsg
+			return err
+		}
```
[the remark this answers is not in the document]
djnakabaale: that's right 😁 , updated

crhchttp/config.go
```diff
@@ -6,19 +6,96 @@
...
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		errMsg := "failed to read response body"
+		log.Error(err, errMsg)
+		return fmt.Errorf("%s: %w", errMsg, err)
+	}
+
+	var result ServiceAccountToken
+	err = json.Unmarshal([]byte(body), &result)
+	if err != nil {
```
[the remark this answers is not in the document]
djnakabaale: updated

crhchttp/config.go
```diff
@@ -6,19 +6,96 @@
...
+
+func (ac *AuthConfig) GetAccessToken(tokenURL string) error {
+	if ac.Authentication != serviceaccount {
+		return nil
+	}
+
+	log := log.WithName("GetAccessToken")
+
+	// Prepare the POST data
+	data := neturl.Values{}
```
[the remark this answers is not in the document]
djnakabaale: updated

crhchttp/config.go
```diff
@@ -6,19 +6,96 @@
...
+	ClientID     string `json:"client_id"`
+	ClientSecret string `json:"client_secret"`
+	GrantType    string `json:"grant_type"`
+}
+type ServiceAccountToken struct {
+	AccessToken      string `json:"access_token"`
+	ExpiresIn        int    `json:"expires_in"`
+	RefreshExpiresIn int    `json:"refresh_expires_in"`
+	TokenType        string `json:"token_type"`
+	NotBeforePolicy  int    `json:"not_before_policy"`
```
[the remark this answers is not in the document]
djnakabaale: updated

crhchttp/config.go
djnakabaale: absolutely, I've updated request to add a timeout

crhchttp/config.go
```diff
@@ -6,19 +6,96 @@
...
+	ServiceAccountToken ServiceAccountToken
+}
+
+// ServiceAccountData provides the data for acquiring the service acount token
+type ServiceAccountData struct {
+	ClientID     string `json:"client_id"`
+	ClientSecret string `json:"client_secret"`
+	GrantType    string `json:"grant_type"`
+}
+type ServiceAccountToken struct {
```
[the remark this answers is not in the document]
djnakabaale: added a docstring

controllers/kokumetricsconfig_controller.go
```diff
@@ -292,6 +292,58 @@ func GetAuthSecret(r *MetricsConfigReconciler, cr *metricscfgv1beta1.MetricsConf
...
+	}
+
+	// Extracting data from the Secret
+	keys := make(map[string]string)
+	for k, v := range secret.Data {
+		keys[strings.ToLower(k)] = string(v)
+	}
+
+	// Defining the required keys
+	requiredKeys := []string{"clientid", "clientsecret"}
```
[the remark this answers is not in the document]
djnakabaale: added a helper function to sanitize the keys

controllers/kokumetricsconfig_controller.go
```diff
@@ -292,6 +292,58 @@ func GetAuthSecret(r *MetricsConfigReconciler, cr *metricscfgv1beta1.MetricsConf
...
+			msg := fmt.Sprintf("service account secret not found with expected %s data", k)
+			log.Info(msg)
+			return fmt.Errorf(msg)
+		}
+	}
+
+	// Populating the authConfig object
+	authConfig.ServiceAccountData = crhchttp.ServiceAccountData{
+		ClientID:     strings.TrimSpace(keys["clientid"]),
+		ClientSecret: strings.TrimSpace(keys["clientsecret"]),
```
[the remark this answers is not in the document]
djnakabaale: removed `strings.TrimSpace` to take secrets as entered by the user

crhchttp/config.go
```diff
@@ -87,10 +86,11 @@ func (ac *AuthConfig) GetAccessToken(cxt context.Context, tokenURL string) error
...
 	body, err := io.ReadAll(resp.Body)
 
+	// TODO: handle errors for ServiceAccountData fields
 	// ONLY proceed to unmarshal if the status was 200
-	if resp.StatusCode != http.StatusOK {
-		return fmt.Errorf("status: %d | error response: %s", resp.StatusCode, body)
-	}
+	// if resp.StatusCode != http.StatusOK {
+	// 	return fmt.Errorf("status: %d | error response: %s", resp.StatusCode, body)
+	// }
```
djnakabaale, line 93: **WIP**: Handle errors specifically due invalid or missing (client-id, client-secret, grant-type) and provide an appropriate error message back to the user

crhchttp/config.go
```diff
@@ -87,10 +86,11 @@ func (ac *AuthConfig) GetAccessToken(cxt context.Context, tokenURL string) error
 
 	body, err := io.ReadAll(resp.Body)
 
+	// TODO: handle errors for ServiceAccountData fields
 	// ONLY proceed to unmarshal if the status was 200
```
djnakabaale, line 90: **WIP**: Handle errors specifically due invalid or missing (client-id, client-secret, grant-type) and provide an appropriate error message back to the user

controllers/kokumetricsconfig_controller.go
```diff
@@ -50,6 +50,8 @@ var (
 	pullSecretAuthKey        = "cloud.openshift.com"
 	authSecretUserKey        = "username"
 	authSecretPasswordKey    = "password"
+	authClientId             = "clientid"
+	authClientSecret         = "clientsecret"
```
maskarb, line 54: Thinking about this some more, when we make the curl request, we're using `client_id` and `client_secret`. I think we should use those here and make customers create their secret using `client_id` and `client_secret` to match how a curl request would be done.

controllers/kokumetricsconfig_controller.go
```diff
@@ -292,6 +305,58 @@ func GetAuthSecret(r *MetricsConfigReconciler, cr *metricscfgv1beta1.MetricsConf
...
+	keys := make(map[string]string)
+	for k, v := range secret.Data {
+		sanitizedKey := sanitizeKey(k)
+		keys[sanitizedKey] = string(v)
+	}
+
+	// Defining the required keys
+	requiredKeys := []string{authClientId, authClientSecret}
+	for _, requiredKey := range requiredKeys {
+		if value, exists := keys[requiredKey]; !exists || value == "" {
```
maskarb, line 343: This is a lot of logic when you can just do
```suggestion
		if len(keys[requiredKey]) <= 0 {
```

https://go.dev/play/p/J0JXP9LikO2

crhchttp/config.go
```diff
@@ -6,19 +6,134 @@
 package crhchttp
```
maskarb, line 6: The permissions on this file changed:
```
100644 → 100755
```
Let's change this back to 644.

crhchttp/config.go
```diff
@@ -6,19 +6,134 @@
...
+			StatusCode:  resp.StatusCode,
+			ErrorType:   errResponse.Error,
+			Description: errResponse.ErrorDescription,
+		}
+	}
+
+	var result ServiceAccountToken
+	if err := json.Unmarshal([]byte(body), &result); err != nil {
+		return fmt.Errorf("failed to unmarshal response: %w", err)
+	}
```
maskarb, line 127: A few things,
1. let's use `json.NewDecoder(resp.Body).Decode(&target)` instead of `io.ReadAll` (I'm not sure where I had gotten io.readall)
2. let's go thru the error pathway for every response that is not a 200. I think we can assume that if the response is not a 200, then the response is not what we need.
3. Let's also use the `AuthError` directly instead of creating a one off `errResponse` struct.
```
type AuthError struct {
	StatusCode       int
	ErrorName        string `json:"error"`
	ErrorDescription string `json:"error_description"`
}

// Error returns a formated error string combining the HTTP status, error type and description.
func (e AuthError) Error() string {
	return fmt.Sprintf("HTTP Status: %d, Error: %s, Description: %s", e.StatusCode, e.ErrorName, e.ErrorDescription)
}
```

```suggestion
	// Handle client errors and invalid requests
	if resp.StatusCode != 200 {
		errResponse := AuthError{StatusCode: resp.StatusCode}

		if err := json.NewDecoder(resp.Body).Decode(&errResponse); err != nil {
			// return generic error with addiotional context if unmarshalling fails
			return fmt.Errorf("status: %d, failed to unmarshal error response: %w", resp.StatusCode, err)
		}
		return errResponse
	}

	var result ServiceAccountToken
	if err := json.NewDecoder(resp.Body).Decode(&result); err != nil {
		return fmt.Errorf("failed to unmarshal response: %w", err)
	}
```

crhchttp/config.go
```diff
@@ -6,19 +6,134 @@
...
+
+	if result.AccessToken == "" {
+		errorMsg := "token response did not contain an access token"
+		log.Error(errors.New(errorMsg), errorMsg)
+		return errors.New(errorMsg)
+	}
```
maskarb, line 133: Is this reasonable to check? This feels like something we shouldn't need to worry about.

controllers/kokumetricsconfig_controller.go
```diff
@@ -347,23 +428,52 @@ func setAuthentication(r *MetricsConfigReconciler, authConfig *crhchttp.AuthConf
...
+					log.Error(authErr, "missing or invalid grant_type in request")
+					specificErrMsg = defaultMsg
+				default:
+					specificErrMsg = defaultMsg
+				}
+				log.Info(specificErrMsg, authErr)
+			}
+			errorMsg := fmt.Sprintf("%s in `%s`. Updated credentials will be re-verified during the next reconciliation.", specificErrMsg, cr.Spec.Authentication.AuthenticationSecretName)
+			cr.Status.Authentication.AuthErrorMessage = errorMsg
+			return err
```
maskarb, line 472: I think all of this is a bit excessive. I would just pass the error we get back from trying to get the token. The error_description we get back is sufficient to pass along. Like this:

```suggestion
			errorMsg := fmt.Sprintf("failed to acquire service-account token: %w" err)
			cr.Status.Authentication.AuthErrorMessage = errorMsg
			return err
```

maskarb (commented):
Some more feedback:

controllers/kokumetricsconfig_controller.go
```diff
@@ -687,6 +687,14 @@ func configurePVC(r *MetricsConfigReconciler, req ctrl.Request, cr *metricscfgv1
...
 func (r *MetricsConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 
+	// TODO: Remove before merge
+	operatorEnv, exists := os.LookupEnv("OPERATOR_RUNTIME_ENV")
+	if exists && operatorEnv == "development" {
+		r.overrideSecretPath = trueDef
+		msg := fmt.Sprintf("using %s environment", operatorEnv)
+		log.Info(msg)
+	}
+
```
djnakabaale, line 697: We can add this in a separate PR If its desirable

crhchttp/http_cloud_dot_redhat.go
```diff
@@ -34,7 +35,14 @@ var log = logr.Log.WithName("crc_http")
...
 var DefaultTransport = &http.Transport{
-	Proxy: http.ProxyFromEnvironment,
+	// TODO: remove conditional proxy if not necessary going forward
+	Proxy: func(req *http.Request) (*url.URL, error) {
+		env := os.Getenv("API_TARGET_ENV")
+		if env == "stage" {
+			return url.Parse("http://squid.corp.redhat.com:3128")
+		}
+		return http.ProxyFromEnvironment(req)
+	},
```
djnakabaale, line 45: This can also be added in a separate PR. Leaving it here for now to get through testing in stage.

api/v1beta1/metricsconfig_types.go
```diff
@@ -98,13 +101,20 @@ type AuthenticationSpec struct {
 	// AuthType is a field of KokuMetricsConfig to represent the authentication type to be used basic or token.
 	// Valid values are:
 	// - "basic" : Enables authentication using user and password from authentication secret.
+	// - "service_account" : Enables authentication using client-id and client-secret from the secret containing service account information.
```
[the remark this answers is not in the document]
djnakabaale: 🦅   👁️ 
nice catch

controllers/kokumetricsconfig_controller.go
```diff
@@ -386,7 +479,8 @@ func validateCredentials(r *MetricsConfigReconciler, handler *sources.SourceHand
 	cr.Status.Authentication.LastVerificationTime = &previousValidation.timestamp
 
 	if err != nil && strings.Contains(err.Error(), "401") {
-		msg := fmt.Sprintf("console.redhat.com credentials are invalid. Correct the username/password in `%s`. Updated credentials will be re-verified during the next reconciliation.", cr.Spec.Authentication.AuthenticationSecretName)
+		// TODO: check with ux/docs on message
+		msg := fmt.Sprintf("console.redhat.com credentials in `%s` are invalid. Basic authentication is being deprecated and will soon no longer be functional. Please switch to service account authentication. Refer to the documentation for more details on this transition.", cr.Spec.Authentication.AuthenticationSecretName)
```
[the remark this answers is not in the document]
djnakabaale: pull this one out

controllers/kokumetricsconfig_controller.go
djnakabaale: 🤦🏾 indeed, I've updated it

crhchttp/config.go
djnakabaale: Nope 🙅🏾 🔕 🚳

controllers/kokumetricsconfig_controller.go
djnakabaale: that sounds good to me. I've cleaned this up.

controllers/kokumetricsconfig_controller.go
djnakabaale: sounds reasonable 👍🏾

maskarb (approved):

PR #212, opened 2023-10-11
update kubebuilder and operator-sdk version

maskarb:
* update the repo structure for the latest kubebuilder (go/v4)
* rely on operator-sdk v1.33
* add basic readiness/liveness probes


samdoran (approved):
It may be a good idea to create a `requirements.txt` file in `scripts/` since `packaging` is required. It would just be the one requirement of `packaging==23.2`.

samdoran (approved):

PR #220, opened 2023-10-27
ppc64le/multi-arch support for operator image

jcho02:
IBM Power (ppc64le) and multi-arch build for the operator image.
Also including ppc64le support tag to the CSV file for operator appear on the OperatorHub in ppc64le arch based OCP clusters.


Makefile
```diff
@@ -166,7 +166,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 # More info: https://docs.docker.com/develop/develop-images/build_enhancements/
 .PHONY: docker-build
 docker-build: ## Build docker image with the manager.
-	$(CONTAINER_TOOL) build -t ${IMG} .
+	$(CONTAINER_TOOL) buildx build --push --platform "linux/amd64,linux/ppc64le" -t ${IMG} .
```
maskarb, line 169: I don't think we need to update this command. We have the `docker-buildx` command below which can be used like:
```
make docker-buildx PLATFORMS='linux/amd64,linux/ppc64le'
```

koku-metrics-operator/3.0.1/manifests/koku-metrics-operator.clusterserviceversion.yaml
```diff
@@ -51,6 +51,10 @@ metadata:
     support: Cost Management
   name: koku-metrics-operator.v3.0.1
   namespace: placeholder
+  labels:
+    operatorframework.io/arch.amd64: supported
+    operatorframework.io/arch.ppc64le: supported
+    operatorframework.io/os.linux: supported
```
maskarb, line 57: This file should not be updated because this is already a published version of the operator. (We will not be updating old versions of the operator to support new architectures)

We should add these here instead:
https://github.com/project-koku/koku-metrics-operator/blob/main/config/manifests/bases/koku-metrics-operator.clusterserviceversion.yaml#L3
so that future versions will contain these labels.

maskarb (changes requested):

jcho02:
Thank you Michael for the feedback, made changes accordingly to your comments

maskarb:
@jcho02
This PR looks totally fine to me, but before we merge it, we need to figure out how we are going to actually build the multi-arch images. The upstream images are currently built in quay.io when PRs are merged and releases are published. It doesn't seem like the quay.io runners support multi-arch image building, so we will need to switch to something else (like a Github action) to accomplish that.

jcho02:
Added a workflow to build and push images onto quay

maskarb:
@jcho02 I went ahead and implemented a build/push pipeline in this PR: https://github.com/project-koku/koku-metrics-operator/pull/231, so feel free to back those changes out of this PR.

If you could run the latest image (`quay.io/project-koku/koku-metrics-operator:latest` or `quay.io/project-koku/koku-metrics-operator@sha256:623a054331239a17c8aaeaa906a7ee7407a6e4ea906b24c1f2c706351b804b4b`) on Power to confirm it compiles correctly for that arch, that would be great.

jcho02:
Thanks @maskarb I verified the image and the operator builds on power platforms. I closed this PR and opened a new one after updating my branch.

config/manifests/bases/koku-metrics-operator.clusterserviceversion.yaml
```diff
@@ -18,6 +18,11 @@ metadata:
     support: Cost Management
   name: koku-metrics-operator.v0.0.0
   namespace: placeholder
+  labels:
+    operatorframework.io/arch.amd64: supported
+    operatorframework.io/arch.ppc64le: supported
```
maskarb, line 23:
```suggestion
    operatorframework.io/arch.arm64: supported
    operatorframework.io/arch.ppc64le: supported
```

Let's also add arm64 here.

maskarb (approved):

PR #221, opened 2023-11-08
[COST-4418] safely convert provider-id to string

maskarb:
[COST-4418](https://issues.redhat.com/browse/COST-4418)


djnakabaale (approved):
lgtm 👍🏾

PR #272, opened 2024-01-18
add node affinity for multiarch clusters

maskarb:
This PR adds node affinity to ensure the controller is added to a node with a supported architecture. 

Reference: https://sdk.operatorframework.io/docs/advanced-topics/multi-arch/#setting-node-affinity-in-a-kubernetes-manifest


djnakabaale (approved):
lgtm 👍🏾

PR #278, opened 2024-01-23
Run github workflows CI jobs in ppc64le

jcho02:
Runs tests in a ppc64le platform by first creating a ppc64le docker container and then running the tests as a non-root user inside the container.
Kubebuilder package for Power only provides the kubebuilder binary so etcd, kube-apiserver, and kubectl binaries are also installed and then mounted onto the docker container.


maskarb:
This is now covered by: https://github.com/project-koku/koku-metrics-operator/pull/279

PR #145, opened 2022-07-05
remove pod affinity

maskarb:

PR #280, opened 2024-01-30
Resolve deployment upgrade with RWO PVCs

montaguethomas:
When OLM tries to upgrade the operator, it will fail to rollout the new pod for the deployment due to RWO volume mounts.

The #141 PR resolved this by setting a preference to schedule the new pod on the name node as the existing pod. However this change was reverted in #145 for some unknown/undocumented reason.

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -19,6 +19,8 @@ spec:
   selector:
     matchLabels:
       control-plane: controller-manager
+  strategy:
+    type: Recreate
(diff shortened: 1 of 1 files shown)


maskarb:
The pod affinities in #141 caused attach errors which is why it was reverted.

We can merge this and try to release it with the 3.2.0 version.

maskarb (approved):

maskarb:
Thanks for the PR!

PR #296, opened 2024-02-20
[COST-4624] do not show progress update during initial data gather

maskarb:
Updating the status during initial data gather is causing the operator to generate invalid reports.

Here is the scenario:
1. a cluster was created in the afternoon on 2-19.
2. the operator was installed on 2-20.
3. During initial data gathering, we start gathering data from 14 days ago. As we progress thru the days with no data, we set the status with the last successful query time and the message that no data was collected. Each of these status updates causes another reconciliation. 
4. While we progress thru the first reconciliation, we [do not gather the data on the 19th because it is incomplete data](https://github.com/project-koku/koku-metrics-operator/blob/main/internal/controller/kokumetricsconfig_controller.go#L829-L836).
5. We gather all of the 2-20 data and ship the payload. This data file remains on the PVC because we have not collected all of 2-20 data yet.
6. The next reconciliation starts (one of the ones that was triggered by the progress update). Let's suppose that during this reconciliation, the status was set to:
```
"last_hour_queried":"2024-02-18 00:00:00 - 2024-02-18 00:59:59","data_collection_message":"No data to report for the hour queried."
```
Because of [this logic](https://github.com/project-koku/koku-metrics-operator/blob/main/internal/controller/prometheus.go#L107), we restart gathering data from 2-18 01:00:00. This causes us to insert the 2-19 data into the report file. So now we have 2-19 data after the 2-20 data in the report files.

The simplest solution for this is to not do the status update until the full reconciliation is done.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -145,11 +145,11 @@ verify-manifests: ## Verify manifests are up to date.
 ENVTEST_K8S_VERSION = 1.28.0
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test ./... -coverprofile cover.out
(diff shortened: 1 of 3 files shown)


djnakabaale (approved):
lgtm 👍🏾

PR #318, opened 2024-03-19
fix the name of the csv

maskarb:
The downstream version of the operator does not include the `v` in the ClusterServiceVersion, so this PR removes it.


djnakabaale (approved):
lgtm 👍🏾

PR #346, opened 2024-05-07
[COST-4115] Add basic auth deprecation notice and transition guidance

djnakabaale:
[COST-4115](https://issues.redhat.com/browse/COST-4115)

This change will add an updated deprecation notice and guidance to transition off basic auth.

### DIFF
--- a/internal/controller/kokumetricsconfig_controller.go
+++ b/internal/controller/kokumetricsconfig_controller.go
@@ -411,8 +411,7 @@ func (r *MetricsConfigReconciler) setAuthentication(ctx context.Context, authCon
 
 	if cr.Status.Authentication.AuthType == metricscfgv1beta1.Basic {
 		// set basic auth deprecation message
-		deprecationMsg := "Basic authentication is deprecated and will be removed after December 31, 2024. Please switch to service-account authentication."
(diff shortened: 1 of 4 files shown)


internal/controller/kokumetricsconfig_controller.go
```diff
@@ -409,6 +410,10 @@ func (r *MetricsConfigReconciler) setAuthentication(ctx context.Context, authCon
 	}
 
 	if cr.Status.Authentication.AuthType == metricscfgv1beta1.Basic {
+		// set basic auth deprecation message
+		deprecationMsg := "Basic authentication is deprecated and will be removed after December 31, 2024. Please switch to service-account authentication."
+		cr.Status.Authentication.DeprecationNotice = deprecationMsg
```
maskarb, line 415: This is likely compiled away, but there really isn't any point in putting this string into a variable:
```suggestion
		cr.Status.Authentication.DeprecationNotice = "Basic authentication is deprecated and will be removed after December 31, 2024. Please switch to service-account authentication."
```

djnakabaale: that's a fair point

PR #370, opened 2024-07-02
[COST-5183] dont use csvReader.ReadAll to improve memory usage

maskarb:
[COST-5183](https://issues.redhat.com/browse/COST-5183)

using ReadAll reads an entire file into memory which is extremely inefficient when we only need the first and last row of the csv. This PR will read line by line to get the first and last rows, which may be slower than reading the whole thing, but will be much more memory efficient and reducing the risk of OOKilling the operator when reading very large files.


djnakabaale (approved):
lgtm 👍🏾

PR #429, opened 2024-09-11
[COST-5382] Konflux - prepare to add operator bundle component

djnakabaale:


maskarb:
This looks ok to me. One other thing I was thinking was: why don't we go ahead and remove the `koku-metrics-operator` directory too? That dir doesn't really serve any purpose. The actual bundle lives in the community-operators-repo, and what we store here is just a copy. It just seems redundant. And, we delete that dir every time we create the downstream branch, so 🤷🏻. It would also be one less thing for us to update during a release.

djnakabaale:
> This looks ok to me. One other thing I was thinking was: why don't we go ahead and remove the `koku-metrics-operator` directory too? That dir doesn't really serve any purpose. The actual bundle lives in the community-operators-repo, and what we store here is just a copy. It just seems redundant. And, we delete that dir every time we create the downstream branch, so 🤷🏻. It would also be one less thing for us to update during a release.

That sounds good to me 👍🏾, I've drop that folder. Thanks giving this a look.

maskarb (approved):

PR #434, opened 2024-09-14
[COST-5382] konflux: update bundle pipeline and use brew registry

djnakabaale:
- update operator bundle pipeline to only run when bundle assets are updated

### DIFF
--- a/.tekton/koku-metrics-operator-bundle-pull-request.yaml
+++ b/.tekton/koku-metrics-operator-bundle-pull-request.yaml
@@ -7,8 +7,13 @@ metadata:
     build.appstudio.redhat.com/pull_request_number: '{{pull_request_number}}'
     build.appstudio.redhat.com/target_branch: '{{target_branch}}'
     pipelinesascode.tekton.dev/max-keep-runs: "3"
-    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch
-      == "main"
+    pipelinesascode.tekton.dev/on-cel-expression: event == "pull" && target_branch == "main" && 
+      (
+        ".tekton/koku-metrics-operator-bundle-pull.yaml".pathChanged() || 
+        "bundle/metadata/***".pathChanged() || 
+        "bundle/manifests/***".pathChanged() || 
+        "bundle.Dockerfile".pathChanged() 
+      )
   creationTimestamp: null
   labels:
     appstudio.openshift.io/application: koku-metrics-operator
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,5 +1,6 @@
 # Build the manager binary
-FROM --platform=${BUILDPLATFORM:-linux/amd64} docker.io/library/golang:1.22.6 AS builder
+FROM --platform=${BUILDPLATFORM:-linux/amd64} brew.registry.redhat.io/rh-osbs/openshift-golang-builder:v1.22 AS builder
```
samdoran, line 2: The base image for upstream should not change. This should be updated during release.

.tekton/koku-metrics-operator-bundle-pull-request.yaml
```diff
@@ -7,8 +7,13 @@ metadata:
...
     pipelinesascode.tekton.dev/max-keep-runs: "3"
-    pipelinesascode.tekton.dev/on-cel-expression: event == "pull_request" && target_branch
-      == "main"
+    pipelinesascode.tekton.dev/on-cel-expression: event == "pull" && target_branch == "main" && 
+      (
+        ".tekton/koku-metrics-operator-bundle-pull.yaml".pathChanged() || 
+        "bundle/metadata/***".pathChanged() || 
+        "bundle/manifests/***".pathChanged() || 
+        "bundle.Dockerfile".pathChanged() 
+      )
```
samdoran, line 16: What's the purpose of changing the trigger for the pipeline? I usually see this used when setting up nudges, but I don't think we have components that need nudging.

Dockerfile
samdoran: Everyone seems to solve this differently. We could maintain separate containerfiles for Konfux and use that in the pipeline params, or we could use `BUILD_ARGS` to pass in a variable that sets the base image.

.tekton/koku-metrics-operator-bundle-pull-request.yaml
djnakabaale: This pipeline doesn't need to run every time but only when bundle assets are updated. In the cluster service yaml in the manifests folder, we reference the [operator image](https://github.com/project-koku/koku-metrics-operator/blob/0ec274a9e1c7f3b581018dbf29870348156761a1/bundle/manifests/koku-metrics-operator.clusterserviceversion.yaml#L42) which will be replaced with the konflux-built image and once everything is configured, that's what will be nudged.

Dockerfile
djnakabaale: Right, I think we can leverage separate [dockerfiles in the pipeline params](https://github.com/project-koku/koku-metrics-operator/blob/0ec274a9e1c7f3b581018dbf29870348156761a1/.tekton/koku-metrics-operator-pull-request.yaml#L29).  But we can revisit this once we have downstream builds.

samdoran (approved):

PR #509, opened 2024-12-05
codecov excludes

maskarb:


djnakabaale (approved):

djnakabaale:
paths to exclude added in https://github.com/project-koku/koku-metrics-operator/pull/558

PR #709, opened 2025-09-25
[COST-1385] Do not upload operator reports without a valid integration

djnakabaale:


bacciotti:
I tried to refactor the test suite a bit. I added a [router pattern](https://github.com/project-koku/koku-metrics-operator/pull/709/files#diff-f7d7a9a5a676515f82d0e398a01962d41c18d9d50112e1852a9edb3a31976bd3R169) to make things more readable and easier to maintain. Some tests were failing due to issues with the URL parsing in the Mock, so to avoid some nested ifs, I introduced this new pattern.
But this is just a suggestion, totally fine to keep it the way it was if you prefer!

internal/controller/costmanagementmetricsconfig_controller_test.go
```diff
@@ -1448,4 +1453,125 @@ var _ = Describe("MetricsConfigController - CRD Handling", Ordered, func() {
...
+			Eventually(func() string {
+				_ = k8sClient.Get(ctx, types.NamespacedName{Name: instCopy.Name, Namespace: namespace}, fetched)
+				return fetched.Status.Upload.UploadError
+			}, timeout, interval).Should(ContainSubstring("Reports are being stored"))
+
+			Expect(fetched.Status.APIURL).To(Equal(defaultAPIURL))
+			Expect(fetched.Status.Upload.LastSuccessfulUploadTime.IsZero()).To(BeTrue())
+		})
+
+		It("should store reports when source validation fails", func() {
```
maskarb, line 1484: Since we're still using the invalid endpoint here, there are no validation errors. So this statement does not align with what this test is checking for.

internal/controller/costmanagementmetricsconfig_controller_test.go
```diff
@@ -1448,4 +1453,125 @@ var _ = Describe("MetricsConfigController - CRD Handling", Ordered, func() {
...
+			createObject(ctx, instCopy)
+
+			fetched := &metricscfgv1beta1.MetricsConfig{}
+
+			Eventually(func() bool {
+				_ = k8sClient.Get(ctx, types.NamespacedName{Name: instCopy.Name, Namespace: namespace}, fetched)
+				return fetched.Status.Authentication.AuthenticationCredentialsFound != nil
+			}, timeout, interval).Should(BeTrue())
+
+			// Should get source validation error and block upload
```
maskarb, line 1497: Again, this test is not configured to throw validation errors, so this statement is not correct.

internal/controller/costmanagementmetricsconfig_controller_test.go
```diff
@@ -1448,4 +1453,125 @@ var _ = Describe("MetricsConfigController - CRD Handling", Ordered, func() {
...
+				_ = k8sClient.Get(ctx, types.NamespacedName{Name: instCopy.Name, Namespace: namespace}, fetched)
+				return fetched.Status.Upload.UploadError
+			}, timeout, interval).Should(ContainSubstring("Reports are being stored"))
+
+			Expect(fetched.Status.APIURL).To(Equal(defaultAPIURL))
+			Expect(fetched.Status.Source.SourceError).ToNot(BeNil())
+			Expect(fetched.Status.Upload.LastSuccessfulUploadTime.IsZero()).To(BeTrue())
+		})
+
+		It("should set generic upload error message when SourceError is empty", func() {
```
maskarb, line 1508: I think this is a leftover test.

internal/controller/costmanagementmetricsconfig_controller_test.go
```diff
@@ -1448,4 +1453,125 @@ var _ = Describe("MetricsConfigController - CRD Handling", Ordered, func() {
...
+			testInstance.Spec.APIURL = validTS.URL
+
+			Expect(k8sClient.Create(ctx, testInstance)).Should(Succeed())
+			defer func() { _ = k8sClient.Delete(ctx, testInstance) }()
+
+			uploadDir := "tmp/koku-metrics-operator-reports/upload"
+			Expect(os.MkdirAll(uploadDir, 0755)).To(Succeed())
+			_, err := Copy(0644, "test_files/testFile.tar.gz", filepath.Join(uploadDir, "test-file.tar.gz"))
+			Expect(err).To(BeNil())
+			defer os.Remove(filepath.Join(uploadDir, "test-file.tar.gz"))
```
maskarb, line 1522: None of this is necessary. We have a helper function to create test files. Use `Expect(setup()).Should(Succeed())` instead.

internal/controller/costmanagementmetricsconfig_controller_test.go
```diff
@@ -1448,4 +1453,125 @@ var _ = Describe("MetricsConfigController - CRD Handling", Ordered, func() {
...
+				return fetched.Status.Source.SourceDefined != nil && *fetched.Status.Source.SourceDefined
+			}, timeout, interval).Should(BeTrue())
+
+			uploadDir := "tmp/koku-metrics-operator-reports/upload"
+			Expect(os.MkdirAll(uploadDir, 0755)).To(Succeed())
+			_, err := Copy(0644, "test_files/testFile.tar.gz", filepath.Join(uploadDir, "upload-error-test.tar.gz"))
+			Expect(err).To(BeNil())
+			defer os.Remove(filepath.Join(uploadDir, "upload-error-test.tar.gz"))
+
+			fetched.Status.Source.SourceDefined = &trueValue
```
maskarb, line 1562: This is already asserted here:
https://github.com/project-koku/koku-metrics-operator/pull/709/files#diff-9d72ed93e35b55c1be30934c04b7a38d50f45b67d2c5d57d1bacc5f2a3f0a13aR1553
But either way, we still cannot set any `status` field like this.

internal/controller/costmanagementmetricsconfig_controller_test.go
```diff
@@ -1448,4 +1453,125 @@ var _ = Describe("MetricsConfigController - CRD Handling", Ordered, func() {
...
+			fetched.Status.Packaging.PackagedFiles = []string{"test-file.tar.gz"}
+			Expect(k8sClient.Status().Update(ctx, fetched)).Should(Succeed())
+
+			Eventually(func() string {
+				_ = k8sClient.Get(ctx, types.NamespacedName{Name: testInstance.Name, Namespace: namespace}, fetched)
+				return fetched.Status.Upload.UploadError
+			}, timeout, interval).Should(Equal("Reports are being stored until a valid integration is configured in console.redhat.com"))
+		})
+
+		It("should return error when uploadFiles fails", func() {
```
maskarb, line 1540: I'm not really sure this test is necessary. We already have upload specific tests elsewhere.

internal/controller/costmanagementmetricsconfig_controller_test.go
```diff
@@ -963,11 +962,68 @@ var _ = Describe("MetricsConfigController - CRD Handling", Ordered, func() {
...
+					return fetched.Status.Source.SourceDefined != nil
+				}, timeout, interval).Should(BeTrue())
+
+				Expect(*fetched.Status.Source.SourceDefined).To(BeTrue()) // Source is valid
+
+				// Upload should fail due to expired credentials at upload endpoint
+				Eventually(func() bool {
+					_ = k8sClient.Get(ctx, types.NamespacedName{Name: instCopy.Name, Namespace: namespace}, fetched)
+					return fetched.Status.Upload.UploadError != ""
+				}, timeout, interval).Should(BeTrue())
```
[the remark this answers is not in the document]
bacciotti: the `Eventually` is meant to handle async operations afaik... yes, we can merge both into a single one that waits for the source validation (SourceDefined) and the upload attempt (UploadError). That way, we make sure the processes are completed in the right order.

maskarb (approved):

PR #938, opened 2026-05-07
Fix gpu_pod_uptime to report actual pod uptime and add gpu_pod_utilization

snecklifter:
## Summary

- **Fixes `gpu_pod_uptime`** to report actual wall-clock pod uptime instead of GPU compute engine active time. The metric was derived from `DCGM_FI_PROF_GR_ENGINE_ACTIVE` (a 0-1 utilization gauge), so a pod running 1 hour at 50% GPU activity reported ~1800s instead of 3600s. A new PromQL query normalizes all samples to 1 using `clamp_max(X + 1, 1)`, so `sum * 60` yields true uptime in seconds.
- **Adds `gpu_pod_utilization`** as a new CSV field to retain the raw utilization sum that `gpu_pod_uptime` previously reported.
- Both metrics correctly handle MIG slices via `GPU_I_ID` and `GPU_I_PROFILE` labels.

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] All collector tests pass (`go test ./internal/collector/...`)
- [x] Non-MIG test verifies both uptime (60s) and utilization (0.5) values in CSV
- [x] MIG test verifies both uptime (60s) and utilization (0.75) values in CSV
- [x] Full integration test (`TestGenerateReports`) passes with updated expected CSV

🤖 Generated with [Claude Code](https://claude.com/claude-code)


djnakabaale (approved):
PR LGTM. Codecov is possibly failing because this is from a fork. All actual tests pass.

PR #973, opened 2026-06-08
[COST-7287] Use dynamic API URL in error messages

martinpovolny:
Use dynamic API URL in error messages for on-prem deployments.

When source validation fails, the operator now displays the actual
API URL in error messages instead of hardcoding "console.redhat.com".
This prevents confusion for on-premise deployments that use custom
API URLs.


internal/controller/url_formatting_test.go
```diff
@@ -0,0 +1,34 @@
+//
+// Copyright 2021 Red Hat Inc.
```
djnakabaale: small nit suggestion

```suggestion
// Copyright 2026 Red Hat Inc.
```

djnakabaale (approved):

PR #999, opened 2026-06-23
Add CRD-configurable GPU namespace exclusion

snecklifter:
## Summary

Addresses: [COST-7760](https://issues.redhat.com/browse/COST-7760)

- GPU metrics reports include rows from infrastructure namespaces (e.g. DCGM exporter pods), inflating cost data with non-billable GPU operator overhead
- Adds `exclude_gpu_namespaces` field to `PrometheusSpec` CRD that filters specified namespaces from GPU cost reports at CSV-write time
- Defaults to an empty list so existing customer workflows are unaffected — customers opt-in to excluding specific namespaces

### Example CR configuration
```yaml
prometheus_config:
  exclude_gpu_namespaces:
    - nvidia-gpu-operator
    - another-infra-namespace
```

## Test plan

- [x] `make generate && make manifests` — CRD schema and DeepCopy regenerated
- [x] `make fmt && make lint` — passes
- [x] `make test` — all tests pass including new `TestGenerateCostNvidiaGpuMetricsReport_ExcludeNamespaces` and `TestReflectSpecExcludeGpuNamespaces`
- [ ] Deploy to cluster with GPU workloads and verify excluded namespace rows are absent from GPU reports


snecklifter:
@djnakabaale nudge on this one

djnakabaale:
> @djnakabaale nudge on this one

@snecklifter  thanks for being patient with my slow reply.

So some customers prefer to account for infra-related costs (like GPU operator)  as overhead that gets distributed back to user workloads. Because of that, I'd recommend not excluding any namespace out of the box as a default. That way we don't impact existing customer workflows in this regard.

The configurable field itself is fine, just let each customer provide the namespace(s) they want excluded rather than pre-populating it.

snecklifter:
Actually, I think that is enough for our use case so will close this one and just exclude namespace at query time.

PR #1050, opened 2026-08-05
[COST-7998] Restrict prometheus service_address to in-cluster .svc urls

djnakabaale:
[COST-7998](https://issues.redhat.com/browse/COST-7998)

Runtime check: only `https` URLs with hostname ending in `.svc` or `.svc.cluster.local` are allowed (validated before the SA token is loaded).

### Prerequisites

* OpenShift cluster (CRC is fine) with the `COST-7998` build deployed
* `oc` logged in; project `koku-metrics-operator`

Optional CRC/arm64 image from local e2e: `quay.io/dnakabaa/koku-metrics-operator:v4.4.1-cost7998`

```sh
export IMG=quay.io/dnakabaa/koku-metrics-operator:v4.4.1-cost7998
make deploy IMG="$IMG"
```

### Scenario 1: Reject — external URL

```yaml
apiVersion: costmanagement-metrics-cfg.openshift.io/v1beta1
kind: CostManagementMetricsConfig
metadata:
  name: costmanagementmetricscfg-sample
spec:
  authentication:
    type: token
  upload:
    upload_toggle: false
  prometheus_config:
    collect_previous_data: false
    service_address: https://evil.example.com
    skip_tls_verification: true
  source:
    create_source: false
```

**Expect:** `prometheus_configured=false`; `configuration_error` contains `in-cluster service URL` and `evil.example.com`.

```sh
oc get costmanagementmetricsconfig costmanagementmetricscfg-sample -n koku-metrics-operator -o yaml
oc logs -n koku-metrics-operator deploy/koku-metrics-operator --tail=50 | grep -E 'in-cluster|Reconciler error'
```

### Scenario 2: Allow — default `.svc`

Patch the same CR:

```yaml
spec:
  prometheus_config:
    service_address: https://thanos-querier.openshift-monitoring.svc:9091
    skip_tls_verification: true
```

**Expect:** `prometheus_configured=true`, empty `configuration_error`, `prometheus_connected=true` (when thanos is reachable).

## Summary by Sourcery

Restrict Prometheus service_address configuration to in-cluster HTTPS service URLs and ensure invalid endpoints are rejected before credentials are loaded.

Bug Fixes:
- Prevent configuring Prometheus with external, non-.svc/.svc.cluster.local, or non-HTTPS service_address values by validating the URL at runtime.

Enhancements:
- Use the default in-cluster Prometheus service address in tests and add helper logic to centralize service_address validation.

Tests:
- Add unit and controller tests verifying allowed and rejected Prometheus service_address values, including that external URLs are blocked before any Prometheus queries or token loading occur.

### DIFF
--- a/internal/collector/prometheus.go
+++ b/internal/collector/prometheus.go
@@ -91,8 +91,13 @@ func statusHelper(cr *metricscfgv1beta1.MetricsConfig, status int, err error) {
 
 type PrometheusConfigurationSetter func(ps *metricscfgv1beta1.PrometheusSpec, c *PrometheusCollector) error
 
-// IsAllowedPromSvcAddress reports whether address is an https URL whose
-// hostname is in-cluster Service DNS (.svc or .svc.cluster.local).
+const (
+	thanosQuerierSvcHost             = "thanos-querier.openshift-monitoring.svc"
+	thanosQuerierSvcClusterLocalHost = "thanos-querier.openshift-monitoring.svc.cluster.local"
+)
+
+// IsAllowedPromSvcAddress reports whether address is an https URL for the
+// OpenShift cluster-monitoring thanos-querier Service.
 func IsAllowedPromSvcAddress(address string) bool {
 	u, err := url.Parse(address)
 	if err != nil || u.Host == "" {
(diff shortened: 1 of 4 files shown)


martinpovolny (dismissed):

martinpovolny:
The PR is good. If you would like to improve further, this looks helpful:

`IsAllowedPromSvcAddress` uses strings.HasSuffix which is case-sensitive, but only the scheme is compared case-insensitively (via EqualFold). A URL like `HTTPS://THANOS.MONITORING.SVC:9091` passes the scheme check but fails the suffix check because HasSuffix(".SVC", ".svc") is false.


The suffix check `strings.HasSuffix(host, ".svc")` accepts any hostname ending in `.svc`, including `evil.com.svc` or `evil.com.svc.cluster.local`. Valid Kubernetes service DNS has the form `<service>.<namespace>.svc[.cluster.local]`, but the check doesn't enforce the namespace component.

djnakabaale:
> The PR is good. If you would like to improve further, this looks helpful:
> 
> `IsAllowedPromSvcAddress` uses strings.HasSuffix which is case-sensitive, but only the scheme is compared case-insensitively (via EqualFold). A URL like `HTTPS://THANOS.MONITORING.SVC:9091` passes the scheme check but fails the suffix check because HasSuffix(".SVC", ".svc") is false.
> 
> The suffix check `strings.HasSuffix(host, ".svc")` accepts any hostname ending in `.svc`, including `evil.com.svc` or `evil.com.svc.cluster.local`. Valid Kubernetes service DNS has the form `<service>.<namespace>.svc[.cluster.local]`, but the check doesn't enforce the namespace component.

I agree. The operator is OpenShift native and instead of gating on `.svc` suffixes, I updated to allow only the OpenShift thanos-querier hosts (`thanos-querier.openshift-monitoring.svc` / `.svc.cluster.local`), with case-insensitive host matching.

martinpovolny (approved):

PR #1065, opened 2026-08-31
[COST-8157] Bump ubi-micro and ubi base images (attr security fix)

bacciotti:
## Summary

- Bump `ubi-micro` from `9.8-1782840931` to `9.8-1787778798` (built 2026-08-26)
- Bump `ubi` build stage from `9.8-1784165989` to `9.8-1787634763`

## Context

https://redhat.atlassian.net/browse/COST-8157

The `attr` package is not pinned in `rpms.lock.yaml`; it comes from the `ubi-micro` base image layer. The target `ubi-micro` image was verified via Red Hat Container Catalog to include `libattr-2.6.0-1.el9_8`.


bacciotti:
addressed in #1054

PR #1083, opened 2026-09-17
[COST-8233] Add RPM lockfile minimum EVR regression tests

jvcsizilio:
## Jira

[COST-8233](https://redhat.atlassian.net/browse/COST-8233)

## Summary

- Add `internal/rpms` regression tests to verify `rpms.lock.yaml` pins OpenSSL at or above the RHSA-172172 fixed build (`1:3.5.8-1.el9_8`).
- Complements the MintMaker lockfile refresh merged in #1079; no lockfile or operator functional changes in this PR.
- Downstream-only guardrail to catch future OpenSSL downgrades during RPM lockfile maintenance.

## Test plan

- [x] `make fmt`
- [x] `go test ./internal/rpms/ -v`
- [ ] `make test` (full suite)
- [ ] `make lint`


internal/rpms/lockfile.go
```diff
@@ -0,0 +1,184 @@
...
+		}
+		if lVal > rVal {
+			return 1
+		}
+	}
+
+	return 0
+}
+
+func compareRelease(left, right string) int {
```
bacciotti: `compareRelease` uses plain string order, not RPM release comparison.

Example: `"9.el9_8"` vs `"10.el9_8"` returns that 9 is greater than 10.
If the minimum EVR later moves to a double-digit release, a lockfile still on `-9.el9_8` can pass this check by mistake.

Can we use rpmvercmp (or the same numeric/alpha split RPM uses) for the release field, and add a test for 9 vs 10?

internal/rpms/lockfile_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+	for _, tc := range cases {
+		got := compareEVR(tc.left, tc.right)
+		if got != tc.want {
+			t.Errorf("compareEVR(%q, %q) = %d, want %d", tc.left, tc.right, got, tc.want)
+		}
+	}
+}
+
+func TestCompareRelease(t *testing.T) {
```
bacciotti, line 58: Please add a case like `1:3.5.8-9.el9_8` vs `1:3.5.8-10.el9_8` (want -1).
That would have caught the lexicographic release bug.

bacciotti (changes requested):
Some comments

internal/rpms/lockfile.go
jvcsizilio: Good catch — fixed in 8451472e.

`compareRelease` / `compareVersion` now both use `rpmvercmp`-style segment comparison (numeric segments compare numerically, so `9.el9_8` < `10.el9_8`). Also covers `~` / `^` the way RPM does.

internal/rpms/lockfile_test.go
jvcsizilio: Added — `1:3.5.8-9.el9_8` vs `1:3.5.8-10.el9_8` (and the reverse) in `TestCompareEVR` / `TestCompareRelease`, plus a few more `rpmvercmp` edge cases. Thanks!

bacciotti (approved):

